PR #2, opened 2016-08-03
Changes build instructions to use Catkin.

damonkohler:
Uses eth-asl repos for Ceres.
Adds missing package dependency.


wohe:
LGTM

PR #5, opened 2016-08-04
Fix draw ordering issue of the RViz visualization.

wohe:
Instead of drawing into a offscreen texture and blitting the
result as an overlay (which shows up on top of everything else)
we directly draw to the RViz scene. To get proper X-rays one
should use a 0.5 gray background for now.

Also fixes the configuration to enable 3D mapping from a bag.


damonkohler:
LGTM

PR #9, opened 2016-08-04
Fixes bug not displaying all submaps in RViz.

wohe:
The 'version_' member of a DrawableSubmap is now properly initialized.
The redundant 'submap_id' in the SubmapQuery response is removed.
The missing slot for reset() is added.


damonkohler:
LGTM

PR #10, opened 2016-08-05
Cleanup of the RViz submaps plugin.

wohe:
Uses a more concise name to show in RViz.
Moves updating the z fading out of the SceneManagerListener.
(to be removed in a future PR)
Simplifies SubmapsDisplay::processMessage() and
DrawableSubmap::Transform().

### DIFF
--- a/cartographer_ros/src/submaps_display.cc
+++ b/cartographer_ros/src/submaps_display.cc
@@ -50,7 +50,6 @@ constexpr char kDefaultTrackingFrame[] = "base_link";
 SubmapsDisplay::SubmapsDisplay()
     : scene_manager_listener_([this]() { UpdateMapTexture(); }),
       tf_listener_(tf_buffer_) {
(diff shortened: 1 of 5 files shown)


cartographer_ros/src/submaps_display.cc
```diff
@@ -151,20 +144,28 @@ void SubmapsDisplay::RequestNewSubmaps() {
...
+      for (auto& submap : trajectory) {
+        submap->SetAlpha(transform_stamped.transform.translation.z);
+      }
+    }
+  } catch (const tf2::TransformException& ex) {
+    ROS_WARN("Could not compute submap fading: %s", ex.what());
+  }
+}
+
 void SubmapsDisplay::UpdateMapTexture() {
```
damonkohler: Seems like this should be renamed?

damonkohler:
LGTM

PR #11, opened 2016-08-05
Fix support for 2D SLAM without IMU.

wohe:
Do not subscribe to the IMU topic and expect IMU data
if expect_imu_data is set to false in the Lua configuration.


damonkohler:
LGTM

PR #16, opened 2016-08-08
Rename to local/global map frame.

wohe:
Changes the confusingly named GetOdometryToMapTransform() function
to GetLocalToGlobalTransform(), since it is transforming from the
local SLAM map frame to the global SLAM map frame.

This is following https://github.com/googlecartographer/cartographer/pull/2.


damonkohler:
LGTM

PR #19, opened 2016-08-10
Drop the leading slash in turtlebot.urdf frame_ids.

wohe:
In tf2, frame_ids should not start with a slash:
http://wiki.ros.org/tf2/Migration


damonkohler:
LGTM

PR #23, opened 2016-08-12
Add launch files for running demos.

wohe:
This adds two launch files for running 2D and 3D demos.
A bag filename is expected as an argument and is played
back using simulation time while showing the result in a
preconfigured RViz window.

### DIFF
--- a/cartographer_ros/configuration_files/demo_2d.rviz
+++ b/cartographer_ros/configuration_files/demo_2d.rviz
@@ -1,0 +1,197 @@
+# Copyright 2016 The Cartographer Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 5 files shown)


damonkohler:
LGTM

carl1218:
hi，
  An error occurred while I was run the demo_2d.launch,
error messages：requires the 'bag_filename' arg to be set.
How should I set the bag_filename?
deeply sorry for this shallow question.

rohbotics:
You can set it like this when you launch,
`roslaunch cartographer_ros demo_2d.launch bag_filename:=<PATH TO BAG FILE>`

Here is a link for reference http://wiki.ros.org/roslaunch/Commandline%20Tools#Passing_in_args

PR #24, opened 2016-08-12
Remap in the launch file was the wrong way around

rohbotics:
Remapping are based only on topic names, not whether you publish or subscribe.

So now we replace any call to topic `horizontal_laser_2d` with topic `scan`, instead of the other way around.


rohbotics:
I signed it!

Open Source Robotics Foundation, Inc.

rohbotics:
Oh, looks like the default is now `/scan` and you were remapping away from that

It was correct

PR #25, opened 2016-08-16
Cmakelist fixes

magazino:
Some minor problems were found in the CMakeLists.txt that prohibited installing the node.

First of all, the EXPORTED_TARGETS were added as dependency, following the ros-indigo specification: 
http://docs.ros.org/indigo/api/catkin/html/howto/format1/cpp_msg_dependencies.html
This leads to the messages being built before the code itself.

Secondly, cartographer_node was added as a target, in order to make the program executable using launch files.

Thirdly, the plugin_description.xml was renamed to rviz_plugin_description.xml.
This is required to install the correct file.

Lastly, a typo was fixed where an extra "}" seemed to be dangling at the end of a line.

### DIFF
--- a/cartographer_ros/CMakeLists.txt
+++ b/cartographer_ros/CMakeLists.txt
@@ -95,6 +95,10 @@ target_link_libraries(cartographer_node
   gflags  # TODO(whess): Use or remove gflags_catkin.
 )
 
+add_dependencies(cartographer_node
+  ${catkin_EXPORTED_TARGETS}
+)
+
 add_library(cartographer_rviz_submaps_visualization
   src/drawable_submap.cc
   src/drawable_submap.h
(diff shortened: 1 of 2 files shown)


cartographer_ros/CMakeLists.txt
```diff
@@ -137,17 +149,21 @@ install(DIRECTORY configuration_files/
...
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}
 )
 
 install(DIRECTORY ogre_media/
-  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/ogre_media}
+  DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/ogre_media
 )
+
+install(PROGRAMS
+         scripts/imu_pub.py
```
damonkohler: This should no longer be needed. If you disable the IMU in the configuration, it's no longer subscribed to.

damonkohler:
Thanks for the fixes! I don't actually see the dangling } correction, though. Perhaps that was a merge error?

You'll need to sign the CLA before we can accept this contribution.

damonkohler:
Obsoleted by fixes in #26.

PR #31, opened 2016-08-26
Add support for publishing nav_msgs/OccupanyGrid messages.

wohe:
Adds an option publish_occupancy_grid (disabled by default) which continuously
computes and publishes OccupancyGrid in a background thread. Only the 2D case is
currently supported.

And small code style and typo fixes.

### DIFF
--- a/cartographer_ros/src/cartographer_node_main.cc
+++ b/cartographer_ros/src/cartographer_node_main.cc
@@ -234,7 +234,7 @@ class Node {
 
   void PublishSubmapList(int64 timestamp);
   void PublishPose(int64 timestamp);
-  void PublishOccupancyGrids();
+  void SpinOccupancyGridThreadForever();
 
   // TODO(hrapp): Pull out the common functionality between this and MapWriter
   // into an open sourcable MapWriter.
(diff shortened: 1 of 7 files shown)


rohbotics:
Yay!

damonkohler:
@wohe Please rebase so that Travis picks this up.

cartographer_ros/src/node_constants.h
```diff
@@ -19,7 +19,7 @@
 
 namespace cartographer_ros {
 
-// The topic that the node will subscrbe under.
+// The topic that the node will subscribe under.
```
damonkohler: subscribe to

cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -634,6 +662,82 @@ void Node::PublishPose(const int64 timestamp) {
   last_pose_publish_timestamp_ = timestamp;
 }
 
+void Node::PublishOccupancyGrids() {
```
damonkohler: PublishOccupanyGrid, there's only the one grid. I think using the plural is the wrong way to indicate that this will loop forever. Maybe ContinuouslyPublishOccupanyGrid? Or PublishOccupancyGridForever?

damonkohler:
LGTM

PR #33, opened 2016-08-29
Adds lua parameter for setting transform wait time

### LINKED ISSUE #32: Tf extrapolation warnings with some setups
I've been trying out the system with a simulated spinning lidar and get quite a bunch of tf related warnings when feeding bag data to cartographer_ros. I thus think there are some waitForTransform() calls missing. See console output here: [tf_warnings_spinning_lidar.txt](https://github.com/googlecartographer/cartographer_ros/files/442195/tf_warnings_spinning_lidar.txt). As LIDARs are mounted fixed in most other example setups, the missing waits likely didn't appear there due to fixed joints being forward timestamped by robot_state_publisher per default.

To reproduce:
* Download bag file: https://drive.google.com/drive/folders/0B1hU91jkd7Vwa3lXZVJPbmZUTWs (Video: https://www.youtube.com/watch?v=KIdIpFkSiHE)
* Get spinning lidar setup from https://github.com/skohlbr/cartographer_ros/commit/90bd6a771014f8a2f07a37974a3bcab3647bc6f2
* Run demo (change command below appropriatly)
```
roslaunch cartographer_ros demo_robot_spin_lidar.launch bag_filename:=/home/kohlbrecher/Downloads/2016-08-04-15-17-14.bag 
```
* Note the launch setup uses https://github.com/team-vigir/vigir_lidar_proc/tree/master/vigir_laserscan_to_pointcloud so far,, it would be preferable to get rid of this dependency.

skohlbr:
Fixes #32 by allowing user to specify tf lookup wait time via the `transform_wait_time` lua parameter. Adjusted the existing config files accordingly.

### DIFF
--- a/cartographer_ros/src/cartographer_node_main.cc
+++ b/cartographer_ros/src/cartographer_node_main.cc
@@ -94,7 +94,6 @@ constexpr int64 kTrajectoryId = 0;
 constexpr int kSubscriberQueueSize = 150;
 constexpr int kSubmapPublishPeriodInUts = 300 * 10000ll;  // 300 milliseconds
 constexpr int kPosePublishPeriodInUts = 5 * 10000ll;      // 5 milliseconds
-constexpr double kMaxTransformDelaySeconds = 0.01;
 constexpr double kSensorDataRatesLoggingPeriodSeconds = 15.;
 
 // Unique default topic names. Expected to be remapped as needed.
@@ -248,6 +247,7 @@ class Node {
   double laser_min_range_;
   double laser_max_range_;
   double laser_missing_echo_ray_length_;
+  double transform_wait_time_;
 
   tf2_ros::Buffer tf_buffer_;
   tf2_ros::TransformListener tf_;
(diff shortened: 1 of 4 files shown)


skohlbr:
If the param is not set in the lua config, it probably defaults to 0 currently, so we have to decide if it should be always required in the config, or if a default value should be set somewhere.

cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -248,6 +247,7 @@ class Node {
   double laser_min_range_;
   double laser_max_range_;
   double laser_missing_echo_ray_length_;
+  double transform_wait_time_;
```
damonkohler: A better name might be lookup_transform_timeout_?

skohlbr: Yeah, that's better. Updating.

damonkohler:
If the parameter isn't in the Lua, we crash. Default values are not supported.

skohlbr:
Ok, but the param is part of the options struct and it appears some of those can remain unset (?). At least for the 3D mapping config, `num_lasers_3d` is set, but it appears to be ommitted  in the 2d config, see https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/configuration_files/backpack_3d.lua#L29 and https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/configuration_files/backpack_2d.lua#L29 (vice versa for the `use_multi_echo_laser_scan_2d` setting).

wohe:
Yes, in this case I wanted to make them optional, although I'm not sure this is the best solution. This is implemented by guarding all accesses with HasKey(). GetDouble() will crash if the parameter is not set.

damonkohler:
Could you please rebase to master so that Travis will pick up this PR?

skohlbr:
Done.

PR #47, opened 2016-09-02
Use WallTimer to always publish pose and submaps.

### LINKED ISSUE #44: No submap_lists published if no data flows
The Cartographer node's global optimization can fall behind and if it does not see any more data coming in, it will stop updating the submap lists, though the optimization is still running. 

Maybe also related to #41

SirVer:
This makes the publishing independent from the inflow of sensor data and
is more in line with how the ROS world works.

Fixes #44.

### DIFF
--- a/cartographer_ros/src/cartographer_node_main.cc
+++ b/cartographer_ros/src/cartographer_node_main.cc
@@ -93,8 +93,6 @@ using carto::kalman_filter::PoseCovariance;
 // TODO(hrapp): Support multi trajectory mapping.
 constexpr int64 kTrajectoryId = 0;
 constexpr int kSubscriberQueueSize = 150;
-constexpr int kSubmapPublishPeriodMs = 300;
-constexpr int kPosePublishPeriodInMs = 5;
 constexpr double kSensorDataRatesLoggingPeriodSeconds = 15.;
 
 // Unique default topic names. Expected to be remapped as needed.
@@ -557,11 +555,13 @@ void Node::Initialize() {
   }
 
   wall_timers_.push_back(node_handle_.createWallTimer(
-      ros::WallDuration(kSubmapPublishPeriodMs / 1000.),
+      ros::WallDuration(
+          lua_parameter_dictionary.GetDouble("submap_publish_period_sec")),
       &Node::PublishSubmapList, this));
   wall_timers_.push_back(node_handle_.createWallTimer(
-      ros::WallDuration(kPosePublishPeriodInMs / 1000.), &Node::PublishPose,
-      this));
+      ros::WallDuration(
+          lua_parameter_dictionary.GetDouble("pose_publish_period_sec")),
+      &Node::PublishPose, this));
 }
 
 bool Node::HandleSubmapQuery(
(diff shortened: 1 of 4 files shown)


cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -93,8 +93,8 @@ using carto::kalman_filter::PoseCovariance;
 // TODO(hrapp): Support multi trajectory mapping.
 constexpr int64 kTrajectoryId = 0;
 constexpr int kSubscriberQueueSize = 150;
-constexpr int kSubmapPublishPeriodInUts = 300 * 10000ll;  // 300 milliseconds
-constexpr int kPosePublishPeriodInUts = 5 * 10000ll;      // 5 milliseconds
+constexpr int kSubmapPublishPeriodMs = 300;
```
damonkohler: InMs

cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -553,6 +555,13 @@ void Node::Initialize() {
     occupancy_grid_thread_ =
         std::thread(&Node::SpinOccupancyGridThreadForever, this);
   }
+
+  wall_timers_.push_back(node_handle_.createWallTimer(
+      ros::WallDuration(kSubmapPublishPeriodMs / 1000.),
```
damonkohler: Maybe keep this constants around in seconds then?

cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -93,8 +93,8 @@ using carto::kalman_filter::PoseCovariance;
 // TODO(hrapp): Support multi trajectory mapping.
 constexpr int64 kTrajectoryId = 0;
 constexpr int kSubscriberQueueSize = 150;
-constexpr int kSubmapPublishPeriodInUts = 300 * 10000ll;  // 300 milliseconds
-constexpr int kPosePublishPeriodInUts = 5 * 10000ll;      // 5 milliseconds
+constexpr int kSubmapPublishPeriodMs = 300;
+constexpr int kPosePublishPeriodInMs = 5;
```
damonkohler: This should be configurable in the Lua.

damonkohler:
LGTM

cartographer_ros/src/cartographer_node_main.cc
SirVer: no longer applicable.

cartographer_ros/src/cartographer_node_main.cc
SirVer: moved to lua.

cartographer_ros/src/cartographer_node_main.cc
SirVer: done.

damonkohler:
LGTM

skohlbr:
This actually doesn't work correctly for all use cases. Currently, the transform is repeatedly published, but always with the last pose estimate timestamp. If pose estimates are internally generated at low rate (such as when aggregating spinning LIDAR data and matching only every few seconds or so), this means that the fixed rate publisher republishes the exact same tf data over and over, but as the timestamp does not progress, all system components waiting for tf connectivity stall until the SLAM pose estimate (and the timestamp alongside it) is updated. I'll open a PR with a fix for discussion.

PR #50, opened 2016-09-08
Build cartographer_ros with -O3.

### LINKED ISSUE #41: can't make the good result using your rosbag(from#30)
@damonkohler ,thanks for your rosbag, but I do it as your commit at #30 twice and get different result, not fine.  my rviz display as following picture:
 could you give me some advise to achieve the result as your paper.

 
screenshot from 2016-09-02 10 44 02

screenshot from 2016-09-02 11 12 02

your paper result

wohe:
Most importantly this should fix #41. Before the ROS integration was built
unoptimized which caused performance issues.

### DIFF
--- a/cartographer_ros/CMakeLists.txt
+++ b/cartographer_ros/CMakeLists.txt
@@ -28,7 +28,23 @@ set(PACKAGE_DEPENDENCIES
   tf2_eigen
 )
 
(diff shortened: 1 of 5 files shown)


damonkohler:
LGTM

PR #52, opened 2016-09-08
Fix configuration files.

wohe:
Fixes configuration files to work with the latest code changes.
Also removes the remapping from turtlebot.launch which does not
seem to be useful so far.


damonkohler:
LGTM

PR #53, opened 2016-09-09
Adds publishing of the scan matched point could.

damonkohler:


cartographer_ros/src/msg_conversion.cc
```diff
@@ -151,40 +180,30 @@ sensor_msgs::LaserScan ToLaserScan(
 
 sensor_msgs::PointCloud2 ToPointCloud2Message(
     const int64 timestamp, const string& frame_id,
+    const ::cartographer::sensor::PointCloud& point_cloud) {
```
SirVer: use pcl::toRosMsg instead. We use pcl::fromROSMsg in sensor_data_producer.cc already.

damonkohler: We would first have to convert our point cloud to a PCL point cloud to do that. It doesn't look like PCL has anything equivalent to Eigen's Map, so that would require copying the data an additional time. Doesn't seem worth it.

wohe:
LGTM

PR #54, opened 2016-09-13
Add support for 2D SLAM with a depth camera.

wohe:
This changes the URDF and timeout for the TurtleBot configuration,
and adds a Dockerfile with everything that is needed to run 2D SLAM
with a depth camera. Verified to work on a TurtleBot 2 with an
ASUS Xtion.

And tiny style fixes.


SirVer:
lgtm.

PR #60, opened 2016-09-16
Makes topic names relative.

### LINKED ISSUE #58: Default topics should be relative, not absolute
Currently, the default topics we publish to are all specified with a leading slash. Instead, the slash should be removed so that all topics are namespaced according to the node name. They can then be remapped into the global space in the launch file.

damonkohler:
Fixes #58

### DIFF
--- a/cartographer_ros/configuration_files/demo_2d.rviz
+++ b/cartographer_ros/configuration_files/demo_2d.rviz
@@ -135,8 +135,8 @@ Visualization Manager:
       Enabled: true
       Map frame: map
       Name: Submaps
-      Submap query service: /cartographer/submap_query
-      Topic: /cartographer/submap_list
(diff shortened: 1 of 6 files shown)


cartographer_ros/configuration_files/demo_2d.rviz
```diff
@@ -135,8 +135,8 @@ Visualization Manager:
       Enabled: true
       Map frame: map
       Name: Submaps
-      Submap query service: /cartographer/submap_query
-      Topic: /cartographer/submap_list
+      Submap query service: /submap_query
```
wohe: Also change the corresponding "cartographer/" in submaps_display.cc?

damonkohler: Done.

PR #65, opened 2016-09-19
Transforms odometry into the tracking frame.

damonkohler:


cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -304,8 +304,16 @@ void Node::AddOdometry(int64 timestamp, const string& frame_id,
         Eigen::Matrix3d::Zero(), rotational;
     // clang-format on
   }
-  map_builder_.GetTrajectoryBuilder(kTrajectoryBuilderId)
-      ->AddOdometerPose(time, pose, applied_covariance);
+  try {
+    const Rigid3d sensor_to_tracking =
+        LookupToTrackingTransformOrThrow(time, odometry.child_frame_id);
```
wohe: Since you only use the "child_frame_id": Pass this as "frame_id" and remove the extra frame_id in the SensorData::Odometry struct?

cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -304,8 +304,16 @@ void Node::AddOdometry(int64 timestamp, const string& frame_id,
...
+    map_builder_.GetTrajectoryBuilder(kTrajectoryBuilderId)
+        ->AddOdometerPose(time, sensor_to_tracking * odometry.pose,
+                          applied_covariance);
+  } catch (const tf2::TransformException& ex) {
+    LOG(WARNING) << "Cannot transform " << frame_id << " -> "
```
wohe: This is inconsistent with the lookup code above.

cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -304,8 +304,16 @@ void Node::AddOdometry(int64 timestamp, const string& frame_id,
         Eigen::Matrix3d::Zero(), rotational;
     // clang-format on
   }
-  map_builder_.GetTrajectoryBuilder(kTrajectoryBuilderId)
-      ->AddOdometerPose(time, pose, applied_covariance);
+  try {
+    const Rigid3d sensor_to_tracking =
+        LookupToTrackingTransformOrThrow(time, odometry.child_frame_id);
+    map_builder_.GetTrajectoryBuilder(kTrajectoryBuilderId)
+        ->AddOdometerPose(time, sensor_to_tracking * odometry.pose,
```
wohe: Since odometry.pose is in frame_id according to the documentation, I'd think that the pose is the position of child_frame_id in it, should this be: odometry.pose * sensor_to_tracking.inverse()?

PR #66, opened 2016-09-19
Pulls out RViz support into a separate package.

### LINKED ISSUE #57: Build Cartographer without RViz Plugin
It is a little annoying to have to build the Rviz plugin on embedded arm platforms.

Can there be some kind of build flag to disable it, or even better spin it out into its own package?

damonkohler:
Fixes #57


cartographer_ros/CMakeLists.txt
```diff
@@ -159,19 +127,3 @@ install(DIRECTORY urdf/
 install(DIRECTORY configuration_files/
   DESTINATION ${CATKIN_PACKAGE_SHARE_DESTINATION}/configuration_files/
 )
-
-install(TARGETS
-  cartographer_rviz_submaps_visualization cartographer_node
```
wohe: Retain the installation of cartographer_node?

PR #68, opened 2016-09-26
Do not drop data when falling behind.

wohe:
Dropping data significantly harms quality of SLAM. This change means that
all data will be kept. If the system falls behind only for a short period
of time this is preferable. Publishing of the pose and submaps does not
work in this case and needs improvement.


cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -87,7 +87,7 @@ using carto::kalman_filter::PoseCovariance;
 
 // TODO(hrapp): Support multi trajectory mapping.
 constexpr int64 kTrajectoryBuilderId = 0;
-constexpr int kSubscriberQueueSize = 150;
+constexpr int kInfiniteQueueSize = 0;
```
damonkohler: Please document that this is for our subscribers. Also, what about publishers?

PR #80, opened 2016-10-04
Add support for finishing and writing 2D maps.

wohe:
When receiving a FinishTrajectory call, the Cartographer nodes catches up on
computing loop closures, runs a final optimization, and when in 2D writes out
a PGM map similar to map_saver.

### DIFF
--- a/cartographer_ros/src/cartographer_node_main.cc
+++ b/cartographer_ros/src/cartographer_node_main.cc
@@ -244,8 +244,7 @@ void Node::AddOdometry(const int64 timestamp, const string& frame_id,
         ->AddOdometerPose(time, odometry.pose * sensor_to_tracking.inverse(),
                           applied_covariance);
   } catch (const tf2::TransformException& ex) {
-    LOG(WARNING) << "Cannot transform " << frame_id << " -> "
-                 << options_.tracking_frame << ": " << ex.what();
+    LOG(WARNING) << ex.what();
   }
 }
 
@@ -266,8 +265,7 @@ void Node::AddImu(const int64 timestamp, const string& frame_id,
                      sensor_to_tracking.rotation() *
                          carto::transform::ToEigen(imu.angular_velocity()));
   } catch (const tf2::TransformException& ex) {
-    LOG(WARNING) << "Cannot transform " << frame_id << " -> "
-                 << options_.tracking_frame << ": " << ex.what();
+    LOG(WARNING) << ex.what();
   }
 }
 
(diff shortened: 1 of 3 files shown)


cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -456,6 +466,47 @@ bool Node::HandleSubmapQuery(
   return true;
 }
 
+bool Node::HandleFinishTrajectory(
+    ::cartographer_ros_msgs::FinishTrajectory::Request& request,
+    ::cartographer_ros_msgs::FinishTrajectory::Response& response) {
+  carto::common::MutexLocker lock(&mutex_);
+  map_builder_.sparse_pose_graph()->RunFinalOptimization();
```
damonkohler: You need to call sensor_collator_->FinishTrajectory first.

cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -456,6 +466,49 @@ bool Node::HandleSubmapQuery(
...
+  // TODO(whess): Add multi-trajectory support.
+  sensor_collator_.FinishTrajectory(kTrajectoryBuilderId);
+  map_builder_.sparse_pose_graph()->RunFinalOptimization();
+  // TODO(whess): Write X-rays in 3D.
+  if (options_.map_builder_options.use_trajectory_builder_2d()) {
+    const auto trajectory_nodes =
+        map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
+    if (!trajectory_nodes.empty()) {
+      std::string filename = request.stem + ".pgm";
+      LOG(INFO) << "Saving final map to " << filename << "...";
```
damonkohler: Enclose filename in ''s

cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -456,6 +466,49 @@ bool Node::HandleSubmapQuery(
...
+          if (grid.data[i] >= 0 && grid.data[i] <= 100) {
+            output_file.put((100 - grid.data[i]) * 254 / 100);
+          } else {
+            constexpr uint8_t kUnknownValue = 205;
+            output_file.put(kUnknownValue);
+          }
+        }
+      }
+      output_file.close();
+      CHECK(output_file) << "Writing " << filename << " failed.";
```
damonkohler: Same here.

cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -456,6 +466,49 @@ bool Node::HandleSubmapQuery(
   return true;
 }
 
+bool Node::HandleFinishTrajectory(
+    ::cartographer_ros_msgs::FinishTrajectory::Request& request,
+    ::cartographer_ros_msgs::FinishTrajectory::Response& response) {
+  carto::common::MutexLocker lock(&mutex_);
+  // TODO(whess): Add multi-trajectory support.
+  sensor_collator_.FinishTrajectory(kTrajectoryBuilderId);
```
damonkohler: Have you tested what happens if this is called while sensor data is still coming in? Looking at the code, I would expect any data that arrives after this process begins to be queued up by ROS and then ignored by the collator.

Would be nice to verify that though. A crash would be bad :)

wohe:
PTAL

PR #82, opened 2016-10-05
Fleshes out the ROS API documentation.

damonkohler:


docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+
+The following ROS API is provided by `cartographer_node`_.
+
+Command-line Flags
+==================
+
+.. TODO(damonkoler): Use an options list if it can be made to render nicely.
+
+\-\-configuration_directory
+  First directory in which configuration files are searched, "second is always
```
SirVer: "second? why are there " there?

docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+  the Cartographer installation to allow including files from there.
+
+\-\-configuration_basename
+  Basename, i.e. not containing any directory prefix, of the configuration file.
```
SirVer: ", e.g. backpack_3d.lua"

makes it clear that this is were the lua files are mentioned.

docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Subscribed Topics
+=================
+
+The following range data topics are mutually exclusive.
+
+scan (`sensor_msgs/LaserScan`_)
+  Currently only supported in 2D. If *use_horizontal_laser* is enabled in the
```
SirVer: drop "currently" everywhere?

docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+  Currently only supported in 2D. If *use_horizontal_multi_echo_laser* is
+  enabled in the :doc:`configuration`, this topic will be used as input for
+  SLAM. Only the first echo is used.
+
+points2 (`sensor_msgs/PointCloud2`_)
+  Currently only supported in 3D. If *num_lasers_3d* is set to 1 in the
+  :doc:`configuration`, this topic will be used as input for SLAM. If
+  *num_lasers_3d* is greater than 1, multiple numbered points2 topics (i.e.
+  points2_1, points2_2, points2_3, ...  up to and including *num_lasers_3d*)
+  will be used as inputs for SLAM.
```
SirVer: mention 'scans_per_accumulation' and its interplay here? Not sure if that is excessive, but more lasers always require tweaking that variable at the minimum.

docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+
+The following additional sensor data topics are optional.
+
+imu (`sensor_msgs/Imu`_)
+  Supported in 2D and 3D. This topic will be used as input for SLAM.
```
SirVer: mention that if not provided, we fake an IMU with Z-up.

docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+
+odom (`nav_msgs/Odometry`_)
+  Supported in 2D and 3D. If *use_odometry_data* is enabled in the
+  :doc:`configuration`, this topic will be used as input for SLAM.
+
+Published Topics
+================
+
+map (`nav_msgs/OccupancyGrid`_)
```
SirVer: Add a note that this is discouraged and only meant for interplay with other ROS tools and simplifying getting started. Real users should use submaps

docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+  Currently only supported in 2D. If subscribed to, a background thread will
+  continuously compute and publish the map. Depending on the size of the map, it
+  can take a few seconds between updates.
+
+scan_matched_points2 (`sensor_msgs/PointCloud2`_)
+  Point cloud as it was used for the purpose of scan-to-submap matching.
```
SirVer: This might contain less points than the laser topics, since filtering takes place before scan matching, i.e. these are the points the SLAM algorithm sees.

docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+=======
+ROS API
+=======
+
+The following ROS API is provided by `cartographer_node`_.
+
+Command-line Flags
+==================
+
+.. TODO(damonkoler): Use an options list if it can be made to render nicely.
```
wohe: typo: damonkohler

docs/source/ros_api.rst
wohe: Remove the quotes in front of second?

docs/source/ros_api.rst
wohe: It's also currently not optional but necessary in 3D.

docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Services
+========
+
+submap_query (`cartographer_ros_msgs/SubmapQuery`_)
+  Fetches the requested submap.
+
+finish_trajectory (`cartographer_ros_msgs/FinishTrajectory`_)
+  Finishes the current trajectory by flushing all queued sensor data, running a
+  final optimization, and writing the map to disk.
```
wohe: Mention that this currently causes shutdown of the node?

docs/source/ros_api.rst
```diff
@@ -0,0 +1,117 @@
...
+
+Required TF Transforms
```
wohe: nit: It is usually written as "tf", i.e. all lowercase. Similar below.

docs/source/ros_api.rst
damonkohler: That's not true. We have a node that does this for TurtleBots, but it has nothing to do with this node.

damonkohler:
PTAL

PR #84, opened 2016-10-06
Do not republish unchanged scan matched point clouds.

wohe:
Before, republishing the point clouds at high frequency caused unnecessary
computational load.

This also fixes an issue seen in RViz:
After replaying a bag, point clouds with the same timestamp were continued to
be published. These were never dropped by the PointCloud2 display leading to
excessive memory consumption and bad visualization performance.


cartographer_ros/src/cartographer_node_main.cc
```diff
@@ -166,6 +166,8 @@ class Node {
   ::ros::Publisher submap_list_publisher_;
   ::ros::ServiceServer submap_query_server_;
   ::ros::Publisher scan_matched_point_cloud_publisher_;
+  carto::common::Time last_published_matched_point_cloud_time_ =
```
damonkohler: last_point_cloud_time_?

PR #91, opened 2016-10-10
Directly use Eigen representations of IMU data.

damonkohler:


cartographer_ros/src/sensor_bridge.cc
```diff
@@ -41,7 +41,11 @@ void SensorBridge::AddOdometryMessage(const string& topic,
 void SensorBridge::AddImuMessage(const string& topic,
                                  const sensor_msgs::Imu::ConstPtr& msg) {
   auto sensor_data = ::cartographer::common::make_unique<SensorData>(
-      msg->header.frame_id, ToCartographer(*msg));
+      msg->header.frame_id,
+      SensorData::Imu{{msg->angular_velocity.x, msg->angular_velocity.y,
```
wohe: Can we introduce a conversion function for vectors, similar to ToRigid3d() and ToPoseCovariance() above? ToVector3d()?

cartographer_ros/src/sensor_bridge.cc
```diff
@@ -41,7 +41,11 @@ void SensorBridge::AddOdometryMessage(const string& topic,
 void SensorBridge::AddImuMessage(const string& topic,
                                  const sensor_msgs::Imu::ConstPtr& msg) {
   auto sensor_data = ::cartographer::common::make_unique<SensorData>(
-      msg->header.frame_id, ToCartographer(*msg));
+      msg->header.frame_id,
```
wohe: Should there be a CHECK, that the covariances do not start with -1, i.e. data is there?

damonkohler:
PTAL

PR #107, opened 2016-10-12
[feature] Use pose covariance from odometry message

wsascha:
To me, it's more practical and consistent to directly use the pose covariance within the nav_msgs/Odometry message.


wsascha:
-->signed.

wsascha:
done.

PR #109, opened 2016-10-12
[feature] Use covariances from odometry msgs

wsascha:
To me it's more consistent to directly use pose covariances from nav_msgs/Odometry messages.


damonkohler:
Thanks for the contribution. We purposely do not use the covariance supplied by Odometry messages since they may be asymmetric. If they're symmetric, there's no need to transform them into the target frame, which is difficult to get correct because of how the UKF works.

Also, a constant variance is much easier to tune. @wohe is working on removing the UKF, which is notoriously difficult to tune correctly. Perhaps that change will make it possible to use the covariances coming from the Odometry messages.

I'll close this for now since, at a minimum, this behavior will be optional. The ability to supply constant variances is useful.

PR #112, opened 2016-10-13
Use the simplified ToLaserFan3D function.

wohe:


cartographer_ros/cartographer_ros/sensor_bridge.cc
```diff
@@ -151,7 +151,7 @@ void SensorBridge::HandleLaserScanProto(
     const string& topic, const ::cartographer::common::Time time,
     const string& frame_id,
     const ::cartographer::sensor::proto::LaserScan& laser_scan) {
-  const auto laser_fan_2d = carto::sensor::ToLaserFan(
```
SirVer: is the function still used somewhere else?

wohe: Before two conversion functions existed that were used only here. Now there is one. This change is mirroring what was changed in the googlecartographer/cartographer#57.

PR #122, opened 2016-10-17
Removes templating on data type.

damonkohler:


SirVer:
You did not run update_cmakelists.py

damonkohler:
Yep, saw that. However, it's needed on cartographer, not cartographer_ros. Working on it. Strange that Travis didn't catch it...

PR #127, opened 2016-10-19
Write PCD file for 3D runs when finalizing a trajectory.

at-wat:
This commit depends on [PR #79 on cartographer](https://github.com/googlecartographer/cartographer/pull/79).

Now, it outputs both of XRay image and PCD.
It would be better to be selectable which method of output to be used through a node parameter or the service call.


SirVer:
Thanks for your contribution!

This needs updating - we now also write .ply files, so this fits nicely in there.

at-wat:
I'll open a new PR, since it needs rebasing.

PR #137, opened 2016-10-21
Adds Docker caching to Travis build.

damonkohler:


Dockerfile
```diff
@@ -13,6 +13,13 @@
 # limitations under the License.
 
 FROM ros:indigo
+COPY cartographer_ros.rosinstall cartographer_ros/
```
SirVer: add a comment why they are listed independently.

PR #143, opened 2016-10-24
Set map->odom transform timestamp to ros::Time::now() to prevent wait issue

skohlbr:
As briefly mentioned in #47, not updating the timestamp of the map->odom transform published at fixed rate results in components always waiting till a new SLAM pose estimate is produced. For robots performing online mapping and navigation this clearly can be undesirable (at least when using tools such as ROS navigation or rviz). 

To demonstrate, here is a video of running the demo bag file from https://github.com/tu-darmstadt-ros-pkg/cartographer_hector_tracker before the change:
[Alt text](https://www.youtube.com/watch?v=Ltq_PF4DmOA)
As visible, the robot "jumps" every few seconds as only at these snapshots valid transform data to the world frame is available.

After the change in this PR the robot moves smoothly:
[Alt text](https://www.youtube.com/watch?v=Dn6ybWnW77M)

Putting this out here for discussion, perhaps how things are stamped should be configurable.

### DIFF
--- a/cartographer_ros/cartographer_ros/node_main.cc
+++ b/cartographer_ros/cartographer_ros/node_main.cc
@@ -402,7 +402,7 @@ void Node::PublishPoseAndScanMatchedPointCloud(
   const Rigid3d tracking_to_map = local_to_map * tracking_to_local;
 
   geometry_msgs::TransformStamped stamped_transform;
-  stamped_transform.header.stamp = ToRos(last_pose_estimate.time);
+  stamped_transform.header.stamp = ros::Time::now();
 
   const auto published_to_tracking = tf_bridge_.LookupToTracking(
       last_pose_estimate.time, options_.published_frame);


damonkohler:
You could argue that this is a sort of extrapolation where our motion model is "we're standing still." We have an underlying motion model we could be exposing instead that should give better results. Perhaps this is a reasonable first step though? @wohe thoughts?

skohlbr:
In principle two approaches make some sense:
* Publishing the transform only when a new estimate is generated (as was the case prior to the changes of #47)
* Publishing with a fixed rate, updating the transform timestamp so ROS tools do not wait as proposed in this PR.

Current master mixes both, it publishes at fixed rate but with fixed timestamp. As far as tf is concerned republishing same transform with same timestamp repeatedly has no effect, so resulting behavior is the same as just publishing the transform once when a new estimate is generated.

With the changes in this PR, the robot effectively moves according to odometry between SLAM pose estimate updates, which is about the most practical thing one can do barring other information. One could certainly argue that handling odometry, (short term) SLAM pose estimates and graph optimized pose estimates in the ROS world could be improved though :)

damonkohler:
> With the changes in this PR, the robot effectively moves according to odometry between SLAM pose estimate updates ...

That's only because you're providing your own odom frame. In the case where that's not happening, the robot would be standing still. I think we can extrapolate internally better by combining other sensor data (e.g. odometry and IMU) along with our constant velocity model. We just need to expose the internal pose estimate, used to initialize scan matches, between scan matches.

cartographer_ros/cartographer_ros/node_main.cc
```diff
@@ -420,6 +420,7 @@ void Node::PublishPoseAndScanMatchedPointCloud(
       tf_broadcaster_.sendTransform(stamped_transform);
     } else {
       stamped_transform.header.frame_id = options_.map_frame;
+      stamped_transform.header.stamp = ros::Time::now();
```
damonkohler: This should happen on line 405.

skohlbr: Done and rebased.

PR #147, opened 2016-10-25
Fixes error of not finding the header files when using catkin_make

thiagodefreitas:
If using catkin_make, there are errors for all the header files inside *cartographer_ros* and *cartographer_rviz*, such as:

/home/thiago/kinetic_ws/src/cartographer_ros/cartographer_ros/cartographer_ros/time_conversion.cc:17:46: fatal error: cartographer_ros/time_conversion.h: No such file or directory
compilation terminated.
/home/thiago/kinetic_ws/src/cartographer_ros/cartographer_ros/cartographer_ros/ros_log_sink.cc:17:43: fatal error: cartographer_ros/ros_log_sink.h: No such file or directory

Adding the INCLUDE_DIRS to the CMakeLists solves this issue


SirVer:
I cannot reproduce these errors - how do you trigger them?

I ask because I *think* your fix is masking a deeper problem - this should be handled in our functions.cmake file instead. But since I do not see the symptoms I cannot follow the fix.

thiagodefreitas:
@SirVer: This happens for me if I just clone cartographer_ros into my kinetic workspace. 

And run any of the following commands:
catkin_make
catkin_make --only-pkg-with-dep cartographer_ros
catkin_make  --pkg cartographer_ros

FYI: All other packages are building fine on this workspace.

SirVer:
Where should it find the Cartographer includes then? Did you install system wide?

thiagodefreitas:
cartographer itself is installed system wide, similarly to ceres.

SirVer:
@googlebot Please check again.

SirVer:
bot, please recheck CLA

SirVer:
@thiagodefreitas

I verified that your change is required - I also verified that it is probably not sufficient in all cases if ceres and cartographer are system wide installed - but let's get it in!

Sorry for the CLA snafu here - googlebot got confused because github decided to use my private email adress when I updated your branch to master. Since you signed the SLA this is good to go in anyways though. Thanks for the contribution!

PR #162, opened 2016-11-08
Fixes urdfdom_headers version compatibility

at-wat:
On latest urdfdom, urdf::LinkSharedPtr uses C++11's std::shared_ptr instead of boost::shared_ptr.


SirVer:
Thanks for the contribution!

PR #184, opened 2016-11-24
Fix trajectory start/stop bug and some cleanups.

damonkohler:
Also improves Tracker configuration based on real data.

### DIFF
--- a/cartographer_ros/cartographer_ros/node_main.cc
+++ b/cartographer_ros/cartographer_ros/node_main.cc
@@ -315,7 +315,7 @@ bool Node::HandleFinishTrajectory(
   const auto trajectory_nodes =
       map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
   if (trajectory_nodes.empty()) {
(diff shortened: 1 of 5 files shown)


cartographer_ros/cartographer_ros/node_main.cc
```diff
@@ -299,22 +299,29 @@ bool Node::HandleSubmapQuery(
...
+
+  map_builder_.FinishTrajectory(previous_trajectory_id);
   map_builder_.sparse_pose_graph()->RunFinalOptimization();
 
   const auto trajectory_nodes =
       map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
   if (trajectory_nodes.empty()) {
-    LOG(WARNING) << "Map is empty and will not be saved.";
-    return true;
+    LOG(WARNING) << "Empty trajectory, no assets will be written.";
```
wohe: Revert? We always write the full map, i.e. if this is empty there is no data in any of the trajectories.

damonkohler: Done.

PR #194, opened 2016-11-28
Support multiple simultaneous trajectories.

damonkohler:


cartographer_ros/cartographer_ros/node.cc
```diff
@@ -64,11 +64,15 @@ constexpr char kFinishTrajectoryServiceName[] = "finish_trajectory";
...
     : options_(options),
       tf_buffer_(::ros::Duration(kTfBufferCacheTimeInSeconds)),
-      tf_(tf_buffer_) {}
+      tf_(tf_buffer_),
+      map_builder_bridge_(options_) {}
 
 Node::~Node() {
   {
     carto::common::MutexLocker lock(&mutex_);
+    if (trajectory_id_ >= 0) {
```
wohe: Should this be a CHECK as it is always the case after calling `Initialize()`?

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -185,27 +187,31 @@ bool Node::HandleSubmapQuery(
...
 
 bool Node::HandleFinishTrajectory(
     ::cartographer_ros_msgs::FinishTrajectory::Request& request,
-    ::cartographer_ros_msgs::FinishTrajectory::Response& response) {
+    ::cartographer_ros_msgs::FinishTrajectory::Response&) {
   carto::common::MutexLocker lock(&mutex_);
-  return map_builder_bridge_->HandleFinishTrajectory(request, response);
+  const int previous_trajectory_id = trajectory_id_;
+  trajectory_id_ = map_builder_bridge_.AddTrajectory(
+      expected_sensor_ids_, options_.tracking_frame, &tf_buffer_);
```
wohe: Since the bridge is the only user of the `tf_buffer_`: Move it there?
Otherwise, since it never changes: Pass to the constructor?

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -64,11 +64,15 @@ constexpr char kFinishTrajectoryServiceName[] = "finish_trajectory";
 Node::Node(const NodeOptions& options)
     : options_(options),
       tf_buffer_(::ros::Duration(kTfBufferCacheTimeInSeconds)),
-      tf_(tf_buffer_) {}
+      tf_(tf_buffer_),
```
wohe: Unused? Remove!

cartographer_ros/cartographer_ros/node.cc
damonkohler: Done.

cartographer_ros/cartographer_ros/node.cc
damonkohler: Done.

damonkohler:
PTAL

PR #195, opened 2016-11-28
Moves TfBridge into SensorBridge.

damonkohler:


cartographer_ros/cartographer_ros/sensor_bridge.h
```diff
@@ -54,13 +55,15 @@ class SensorBridge {
   void HandlePointCloud2Message(const string& sensor_id,
                                 const sensor_msgs::PointCloud2::ConstPtr& msg);
 
+  const TfBridge& tf_bridge();
```
wohe: Make the function `const` as well?

PR #198, opened 2016-11-29
Pulls out sensor data wiring into main.

damonkohler:


cartographer_ros/cartographer_ros/node.cc
```diff
@@ -49,17 +49,11 @@ constexpr int kInfiniteSubscriberQueueSize = 0;
 constexpr int kLatestOnlyPublisherQueueSize = 1;
```
wohe: Is the `kInfiniteSubscriberQueueSize` not used anymore? Remove!

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -182,25 +100,19 @@ void Node::Initialize() {
       &Node::PublishTrajectoryStates, this));
 }
 
+void Node::SpinForever() { ::ros::spin(); }
```
wohe: Should this be renamed `Spin()` for consistency and also the fact that it will return when the node terminates?

cartographer_ros/cartographer_ros/node.cc
damonkohler: Done.

cartographer_ros/cartographer_ros/node.cc
damonkohler: Done.

PR #204, opened 2016-11-30
Adds multi-trajectory support to offline mapping.

damonkohler:


cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -49,8 +49,18 @@ namespace {
 constexpr int kLatestOnlyPublisherQueueSize = 1;
 constexpr char kClockTopic[] = "clock";
 
-void Run() {
-  // TODO(damonkohler): Pull out this common code across binaries.
+std::vector<string> SplitString(const string& input, const char delimeter) {
```
wohe: typo: `delimiter`

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -49,8 +49,18 @@ namespace {
...
+  std::stringstream stream;
```
wohe: `std::istringstream stream(input);`?
No need to call `str()`.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -106,61 +116,68 @@ void Run() {
...
 
-    rosgraph_msgs::Clock clock;
-    clock.clock = msg.getTime();
-    clock_publisher.publish(clock);
+    for (const rosbag::MessageInstance& msg : rosbag::View(bag)) {
+      if (!::ros::ok()) {
+        return;
+      }
+
+      const string& topic = node.node_handle()->resolveName(msg.getTopic());
```
wohe: `const string topic` to be explicit that `resolveName()` return a value?

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -106,61 +116,68 @@ void Run() {
...
+      if (expected_sensor_ids.count(
+              node.node_handle()->resolveName(msg.getTopic())) == 0) {
```
wohe: Use `topic`?

cartographer_ros/cartographer_ros/offline_node_main.cc
damonkohler: Done.

cartographer_ros/cartographer_ros/offline_node_main.cc
damonkohler: Done.

PR #206, opened 2016-12-01
Cleanups and preparation for handling /tf topic.

damonkohler:


cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -169,6 +173,11 @@ void Run(std::vector<string> bag_filenames) {
       clock_publisher.publish(clock);
 
       ::ros::spinOnce();
+
+      const ::ros::Time time = msg.getTime();
```
wohe: This is used exactly one: Inline?
Same in the other file.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -128,8 +128,12 @@ void Run(std::vector<string> bag_filenames) {
         expected_sensor_ids, options.tracking_frame);
     rosbag::Bag bag;
     bag.open(bag_filename, rosbag::bagmode::Read);
+    rosbag::View view(bag);
 
-    for (const rosbag::MessageInstance& msg : rosbag::View(bag)) {
+    const ::ros::Time begin_time = view.getBeginTime();
+    const ::ros::Time end_time = view.getEndTime();
```
wohe: Since you never use the `end_time`:
`const ::ros::Duration bag_duration = view.getEndTime() - begin_time;`
here?

PR #213, opened 2016-12-05
Optionally read /tf from bag in place of a URDF.

damonkohler:


cartographer_ros/cartographer_ros/assets_writer_main.cc
```diff
@@ -60,68 +62,67 @@ namespace carto = ::cartographer;
...
+  carto::io::RegisterBuiltInPointsProcessors(trajectory_proto, &builder);
   std::vector<std::unique_ptr<carto::io::PointsProcessor>> pipeline =
       builder.CreatePipeline(
           lua_parameter_dictionary.GetDictionary("pipeline").get());
 
-  tf2_ros::Buffer buffer;
-  // TODO(hrapp): Also parse tf messages and keep the 'buffer' updated as to
-  // support non-rigid sensor configurations.
+  auto tf_buffer = ::cartographer::common::make_unique<tf2_ros::Buffer>();
   if (!urdf_filename.empty()) {
```
SirVer: Why not do both if both are specified?

cartographer_ros/cartographer_ros/assets_writer_main.cc
```diff
@@ -60,68 +62,67 @@ namespace carto = ::cartographer;
...
-    }
-    pipeline.back()->Process(std::move(batch));
-  };
-
+  const string tracking_frame =
+      lua_parameter_dictionary.GetString("tracking_frame");
+  const auto transform_interpolation_buffer =
+      carto::transform::TransformInterpolationBuffer::FromTrajectory(
+          trajectory_proto);
+  const auto handle_point_cloud_2_message =
```
SirVer: pull out into a separate method? this gets really long and hard to understand.

cartographer_ros/cartographer_ros/assets_writer_main.cc
damonkohler: The assumption is that, if there's no URDF specified, then all the necessary transforms should be available on the /tf topic. Loading the URDF only makes sense for statically linked sensors (i.e. no movable joints). I fear combining the two approaches would only lead to headaches and I don't see a use case right now.

PR #214, opened 2016-12-06
Include what you use.

SirVer:
I think this should have been caught by travis. Could it be that caching is still messing with us?


damonkohler:
I don't think caching is to blame. You can see it built and linked the binaries successfully before the last PR:

https://api.travis-ci.org/jobs/181579581/log.txt?deansi=true

```docker build . --no-cache``` also works fine for me on my workstation.

PR #219, opened 2016-12-08
Fix offline node.

damonkohler:
Flips topic remapping to match online node and checks that topics are
wired to unique sensor inputs.
Fixes broken launch files.


cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -92,15 +91,17 @@ void Run(std::vector<string> bag_filenames) {
   node.Initialize();
 
   std::unordered_set<string> expected_sensor_ids;
+  const auto check_insert = [&](const string& topic) {
```
SirVer: please make explicit what you capture

damonkohler: Done.

PR #222, opened 2016-12-09
Adds Kinetic to build matrix.

### LINKED ISSUE #176: build fails optimization_problem.cc
Build fails in optimization_problem.cc with "undefined reference to ceres::..."
commit 0c8d6afc33531761e8dd69760907d730a8440ed3 still worked, did not try the commits in between. Also, there seems to be an issue with configuration_files_test.
I am using ROS Kinetic on Ubuntu 16.04LTS.

damonkohler:
Fixes #176

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+**/Dockerfile*
+**/.dockerignore
+**/.git
+**/.travis.yml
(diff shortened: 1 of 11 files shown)


SirVer:
@damonkohler Could you explain what the actual fix was?

damonkohler:
The "fix" was adding the target dependencies before the system dependencies. Not much of a fix as a surprise workaround with the desired effects. The real issue is still TBD.

PR #235, opened 2016-12-20
Add support for making assets out of LaserScan and MultiEchoLaserScan.

SirVer:


cartographer_ros/cartographer_ros/assets_writer_main.cc
```diff
@@ -59,8 +61,31 @@ namespace {
 
 namespace carto = ::cartographer;
 
-void HandlePointCloud2Message(
-    const sensor_msgs::PointCloud2::ConstPtr& msg, const string& tracking_frame,
+carto::sensor::LaserFan ConvertMessage(
+    const sensor_msgs::PointCloud2::ConstPtr& msg) {
```
damonkohler: s/msg/message/g

cartographer_ros/cartographer_ros/assets_writer_main.cc
```diff
@@ -59,8 +61,31 @@ namespace {
...
+  return carto::sensor::ToLaserFan(ToCartographer(*msg));
+}
+
+carto::sensor::LaserFan ConvertMessage(
+    const sensor_msgs::LaserScan::ConstPtr& msg) {
+  return carto::sensor::ToLaserFan(ToCartographer(*msg));
+}
+
+template <typename T>
+void HandleMsg(
```
damonkohler: HandleMessage

cartographer_ros/configuration_files/assets_writer.lua
```diff
@@ -7,11 +7,11 @@ options = {
       -- action = "fixed_ratio_sampler",
       -- sampling_ratio = 1.,
     -- },
-    -- {
-      -- action = "min_max_range_filter",
-      -- min_range = 1.,
-      -- max_range = 150.,
-    -- },
+    {
```
damonkohler: Should this be enabled by default? Also, laser fans contain their min/max range. Maybe that should be respected?

SirVer: It is respected when extracting the laser scan. The hokuyo seem to return some ringing though which is filtered out.

As of having this enabled: I think it is unclear, but in any case for any lasers these are pretty sane values for reasonable looking assets. So I want to keep it for now.

SirVer:
PTAL.

SirVer:
This changed a bit, PTAL.

PR #258, opened 2017-01-25
Fix configuration checks during start up.

wohe:
We now destroy the LuaParameterDictionary directly after
constructing the NodeOptions, so that it is verified
whether all mentioned parameters are understood.
Before, this was checked when shutting down the node.

### DIFF
--- a/cartographer_ros/cartographer_ros/offline_node_main.cc
+++ b/cartographer_ros/cartographer_ros/offline_node_main.cc
@@ -79,7 +79,7 @@ NodeOptions LoadOptions() {
   return CreateNodeOptions(&lua_parameter_dictionary);
 }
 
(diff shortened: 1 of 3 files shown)


cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -76,6 +76,12 @@ void Run(std::vector<string> bag_filenames) {
   cartographer::common::LuaParameterDictionary lua_parameter_dictionary(
       code, std::move(file_resolver));
 
+  return CreateNodeOptions(&lua_parameter_dictionary);
+}
+
+void Run(std::vector<string> bag_filenames) {
```
damonkohler: const ref

PR #266, opened 2017-02-17
Offline node improvements

ojura:
This makes the offline node behave more like a combination of rosbag play + Cartographer online node + robot_state_publisher: 

- enable using both urdf and bag transforms at the same time
- republish transforms from the bag, which partially resolves a TODO for @damonkohler, see the commit messages for more details

### DIFF
--- a/cartographer_ros/cartographer_ros/bag_reader.cc
+++ b/cartographer_ros/cartographer_ros/bag_reader.cc
@@ -29,14 +29,12 @@ namespace cartographer_ros {
 
 constexpr char kTfStaticTopic[] = "/tf_static";
 
-std::unique_ptr<tf2_ros::Buffer> ReadTransformsFromBag(
-    const string& bag_filename) {
+void ReadTransformsFromBag(
+    const string& bag_filename, tf2_ros::Buffer* const buffer) {
   rosbag::Bag bag;
   bag.open(bag_filename, rosbag::bagmode::Read);
   rosbag::View view(bag);
 
-  auto tf_buffer =
-      ::cartographer::common::make_unique<tf2_ros::Buffer>(::ros::DURATION_MAX);
   const ::ros::Time begin_time = view.getBeginTime();
   const double duration_in_seconds = (view.getEndTime() - begin_time).toSec();
   for (const rosbag::MessageInstance& msg : view) {
@@ -45,7 +43,7 @@ std::unique_ptr<tf2_ros::Buffer> ReadTransformsFromBag(
       for (const auto& transform : tf_msg->transforms) {
         try {
           // TODO(damonkohler): Handle topic remapping.
-          tf_buffer->setTransform(transform, "unused_authority",
+          buffer->setTransform(transform, "unused_authority",
                                   msg.getTopic() == kTfStaticTopic);
         } catch (const tf2::TransformException& ex) {
           LOG(WARNING) << ex.what();
@@ -58,7 +56,6 @@ std::unique_ptr<tf2_ros::Buffer> ReadTransformsFromBag(
   }
 
   bag.close();
-  return tf_buffer;
 }
 
 }  // namespace cartographer_ros
--- a/cartographer_ros/cartographer_ros/bag_reader.cc
+++ b/cartographer_ros/cartographer_ros/bag_reader.cc
@@ -30,7 +30,7 @@ namespace cartographer_ros {
 constexpr char kTfStaticTopic[] = "/tf_static";
 
 void ReadTransformsFromBag(
-    const string& bag_filename, tf2_ros::Buffer* const buffer) {
+    const string& bag_filename, tf2_ros::Buffer* const tf_buffer) {
   rosbag::Bag bag;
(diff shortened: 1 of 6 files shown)


cartographer_ros/cartographer_ros/bag_reader.cc
```diff
@@ -29,14 +29,12 @@ namespace cartographer_ros {
 
 constexpr char kTfStaticTopic[] = "/tf_static";
 
-std::unique_ptr<tf2_ros::Buffer> ReadTransformsFromBag(
-    const string& bag_filename) {
+void ReadTransformsFromBag(
+    const string& bag_filename, tf2_ros::Buffer* const buffer) {
```
damonkohler: tf_buffer?

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -139,10 +145,20 @@ void Run(const std::vector<string>& bag_filenames) {
...
+          kTfTopic, kLatestOnlyPublisherQueueSize);
+
+  ::tf2_ros::StaticTransformBroadcaster static_tf_broadcaster;
+
   ::ros::Publisher clock_publisher =
       node.node_handle()->advertise<rosgraph_msgs::Clock>(
           kClockTopic, kLatestOnlyPublisherQueueSize);
 
+  if (urdf_transforms.size() > 0) {
+    static_tf_broadcaster.sendTransform(urdf_transforms);
```
damonkohler: This shouldn't be necessary since the URDF transforms are marked as static.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -162,7 +178,12 @@ void Run(const std::vector<string>& bag_filenames) {
         break;
       }
 
-      // TODO(damonkohler): Republish non-conflicting tf messages.
+      // TODO(damonkohler): Check if republished tf messages are in conflict
```
damonkohler: This behavior needs to be guarded by a flag if conflicts are not handled. Otherwise, it's impossible to replay a bag containing conflicting tf messages.

cartographer_ros/cartographer_ros/bag_reader.cc
ojura: Okay, sure. I was mirroring `ReadStaticTransformsFromUrdf`, where this is also called `buffer`.

cartographer_ros/cartographer_ros/offline_node_main.cc
ojura: Do you mean we can use a regular tf_broadcaster here? There is already one in the class `Node`, but we can't access it from here because it's a private member.

cartographer_ros/cartographer_ros/offline_node_main.cc
ojura: Ok, I will add a flag. What exactly would be a conflicting transform? Is it the following?

- a transformation which has Cartographer's `published_frame` as a child
- in case `provide_odom_frame` is true, a transformation which has `odom_frame` as child

Would it be enough to run a loop through the transforms in the message and do not publish them if they match one of the above, or did you have something else in mind?

cartographer_ros/cartographer_ros/offline_node_main.cc
damonkohler: Sorry, I was confused last time I looked at this. I agree that it makes sense to have this here.

ojura:
One more thing; maybe it would be better to change the flag to also disable loading of the bag transforms into the buffer; in that case, it should be called 'use_bag_transforms' in that case

cartographer_ros/cartographer_ros/offline_node_main.cc
damonkohler: A conflicting transform is one that is both in the recorded bag and will be computed by the system. The logic here would have to duplicate the [logic in Node](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/node.cc#L138) to determine which transforms to filter out.

It's not so difficult to do, but the duplication concerns me as it could lead to issues that may be difficult to debug if the logic ever changes.

Perhaps a better approach is to allow filtering out arbitrary transforms on the command line. We have a small [node that does this already](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/scripts/tf_remove_frames.py).

damonkohler:
I agree, that makes sense.

ojura:
Okay, this should be it. I have also added the same flag to `assets_writer_main`. We should now get the same behavior as before when a .urdf profile is provided and the flag is disabled, while the new default behavior is to use and publish the transformations.

Selective TF frame filtering as you have described can remain a todo, but I think it's okay that, for now, the onus remains on the user to provide a clean bag in case they want to use the bag transformations. Maybe add a utility script similar to the one you mentioned ([example](https://gist.github.com/ojura/99c5ac524b58124ef2da95adb3e46222)), which rewrites the bag without the specified frames, so it can be used with the offline node.

PR #284, opened 2017-03-22
Remove unnecessary dependency on Cartographer.

SirVer:


damonkohler:
Looks like this dependency isn't so unnecessary. Is there a dependency missing in cartographer_ros? Is the dependency graph broken? Travis seems to be producing a believable result.

PR #289, opened 2017-03-28
Rviz plugin improvements

ojura:
This pull request adds the ability to toggle visibility of individual submaps in the rviz plugin. I have also included a fix for a race condition which can happen during disabling of the plug-in (commits f64fe89 and 78583f3), and a few other miscellaneous changes.

### DIFF
--- a/cartographer_ros_msgs/msg/SubmapEntry.msg
+++ b/cartographer_ros_msgs/msg/SubmapEntry.msg
@@ -12,5 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-uint32 submap_version
+int32 submap_version
 geometry_msgs/Pose pose
--- a/cartographer_rviz/cartographer_rviz/drawable_submap.cc
+++ b/cartographer_rviz/cartographer_rviz/drawable_submap.cc
@@ -83,9 +83,14 @@ DrawableSubmap::DrawableSubmap(const int trajectory_id, const int submap_index,
     scene_node_->attachObject(manual_object_);
   }
   connect(this, SIGNAL(RequestSucceeded()), this, SLOT(UpdateSceneNode()));
+  connect(this, SIGNAL(VisibilityChanged(DrawableSubmap*)),
+          this, SLOT(UpdateSceneNode()));
 }
 
 DrawableSubmap::~DrawableSubmap() {
+  ::cartographer::common::MutexLocker locker(&mutex_);
+  quitting_ = true;
+  locker.Await([this]() REQUIRES(mutex_) { return !query_in_progress_; });
   Ogre::MaterialManager::getSingleton().remove(material_->getHandle());
   if (!texture_.isNull()) {
     Ogre::TextureManager::getSingleton().remove(texture_->getHandle());
@@ -128,13 +133,15 @@ void DrawableSubmap::Update(
 
 bool DrawableSubmap::MaybeFetchTexture(ros::ServiceClient* const client) {
   ::cartographer::common::MutexLocker locker(&mutex_);
-  const bool newer_version_available = texture_version_ < metadata_version_;
+  const bool newer_version_available =
+      response_.submap_version < metadata_version_;
   const std::chrono::milliseconds now =
       std::chrono::duration_cast<std::chrono::milliseconds>(
           std::chrono::system_clock::now().time_since_epoch());
   const bool recently_queried =
       last_query_timestamp_ + kMinQueryDelayInMs > now;
-  if (!newer_version_available || recently_queried || query_in_progress_) {
+  if (!newer_version_available || recently_queried
+      || query_in_progress_ || render_in_progress_ || quitting_) {
     return false;
   }
   query_in_progress_ = true;
@@ -148,7 +155,11 @@ bool DrawableSubmap::MaybeFetchTexture(ros::ServiceClient* const client) {
       // 'response_' member to simplify the signal-slot connection slightly.
       ::cartographer::common::MutexLocker locker(&mutex_);
       response_ = srv.response;
-      Q_EMIT RequestSucceeded();
+      query_in_progress_ = false;
+      if (!quitting_) {
+        render_in_progress_ = true;
+        Q_EMIT RequestSucceeded();
+      }
     } else {
       ::cartographer::common::MutexLocker locker(&mutex_);
       query_in_progress_ = false;
@@ -157,9 +168,9 @@ bool DrawableSubmap::MaybeFetchTexture(ros::ServiceClient* const client) {
   return true;
 }
 
-bool DrawableSubmap::QueryInProgress() {
+bool DrawableSubmap::QueryOrRenderInProgress() {
   ::cartographer::common::MutexLocker locker(&mutex_);
-  return query_in_progress_;
+  return query_in_progress_ || render_in_progress_;
 }
 
 void DrawableSubmap::SetAlpha(const double current_tracking_z) {
@@ -176,13 +187,18 @@ void DrawableSubmap::SetAlpha(const double current_tracking_z) {
 
 void DrawableSubmap::UpdateSceneNode() {
   ::cartographer::common::MutexLocker locker(&mutex_);
+  render_in_progress_ = false;
+  if (!GetVisibility()
+      || texture_version_ == response_.submap_version
+      || quitting_) {
+    return;
+  }
   texture_version_ = response_.submap_version;
   std::string compressed_cells(response_.cells.begin(), response_.cells.end());
   std::string cells;
   ::cartographer::common::FastGunzipString(compressed_cells, &cells);
   tf::poseMsgToEigen(response_.slice_pose, slice_pose_);
   UpdateTransform();
-  query_in_progress_ = false;
   // The call to Ogre's loadRawData below does not work with an RG texture,
   // therefore we create an RGB one whose blue channel is always 0.
   std::vector<char> rgb;
--- a/cartographer_rviz/cartographer_rviz/drawable_submap.cc
+++ b/cartographer_rviz/cartographer_rviz/drawable_submap.cc
@@ -25,6 +25,7 @@
 #include "Eigen/Geometry"
 #include "OgreGpuProgramParams.h"
 #include "OgreImage.h"
+#include "cartographer/common/make_unique.h"
 #include "cartographer/common/port.h"
 #include "cartographer_ros_msgs/SubmapQuery.h"
 #include "eigen_conversions/eigen_msg.h"
@@ -54,7 +55,9 @@ std::string GetSubmapIdentifier(const int trajectory_id,
 }  // namespace
 
 DrawableSubmap::DrawableSubmap(const int trajectory_id, const int submap_index,
-                               Ogre::SceneManager* const scene_manager)
+                               Ogre::SceneManager* const scene_manager,
+                               ::rviz::Property* submap_category,
+                               bool initial_visibility)
     : trajectory_id_(trajectory_id),
       submap_index_(submap_index),
       scene_manager_(scene_manager),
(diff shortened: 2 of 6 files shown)


cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -62,19 +65,25 @@ class DrawableSubmap : public QObject {
   bool MaybeFetchTexture(ros::ServiceClient* client);
 
   // Returns whether an RPC is in progress.
-  bool QueryInProgress();
+  bool QueryOrRenderInProgress();
 
   // Sets the alpha of the submap taking into account its slice height and the
   // 'current_tracking_z'.
   void SetAlpha(double current_tracking_z);
+  const int& GetSubmapIndex();
```
damonkohler: Simply return the value here (the reference is confusing and doesn't help) and the method can be const. In the case of a cheap accessor like this, we inline the implementation and use variable-like names:

int submap_index() const { return submap_index_; }

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -62,19 +65,25 @@ class DrawableSubmap : public QObject {
...
+  const int& GetTrajectoryId();
```
damonkohler: Same here.

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -62,19 +65,25 @@ class DrawableSubmap : public QObject {
...
+  bool GetVisibility();
```
damonkohler: Same here.

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -62,19 +65,25 @@ class DrawableSubmap : public QObject {
...
+  void SetVisibility(bool visibility);
```
damonkohler: void set_visibility(bool visibility);

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -46,7 +47,9 @@ class DrawableSubmap : public QObject {
   // Each submap is identified by a 'trajectory_id' plus a 'submap_index'.
   // 'scene_manager' is the Ogre scene manager to which to add a node.
   DrawableSubmap(int trajectory_id, int submap_index,
-                 Ogre::SceneManager* scene_manager);
+                 Ogre::SceneManager* scene_manager,
+                 ::rviz::Property* submap_category,
+                 bool initial_visibility = true);
```
damonkohler: Please don't use default args.

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -54,6 +56,12 @@ SubmapsDisplay::SubmapsDisplay() : tf_listener_(tf_buffer_) {
       "Tracking frame", kDefaultTrackingFrame,
       "Tracking frame, used for fading out submaps.", this);
   client_ = update_nh_.serviceClient<::cartographer_ros_msgs::SubmapQuery>("");
+  submaps_category_ = new ::rviz::Property(
+      "Submaps", QVariant(), "The list of all submaps.", this);
```
damonkohler: For consistency, please drop "The" from help text like this. This becomes, "List of all submaps."

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -54,6 +56,12 @@ SubmapsDisplay::SubmapsDisplay() : tf_listener_(tf_buffer_) {
...
+  visibility_all_enabled_ = new ::rviz::BoolProperty(
+      "All Enabled", true,
+      "Whether all the submaps should be displayed or not.",
+      submaps_category_, SLOT( AllEnabledChanged() ), this);
```
damonkohler: Please run clang-format over your change.

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -104,23 +112,28 @@ void SubmapsDisplay::processMessage(
...
-                trajectory_id, submap_index, context_->getSceneManager()));
+                trajectory_id, submap_index, context_->getSceneManager(),
+                submaps_category_, visibility_all_enabled_->getBool()));
+        ConnectVisibilityChange(trajectory.back());
       }
       trajectory[submap_index]->Update(msg->header,
                                        submap_entries[submap_index],
                                        context_->getFrameManager());
     }
+    if (submap_entries.size() < trajectory.size()) {
```
damonkohler: How does this happen? Please add a comment.

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -104,23 +112,28 @@ void SubmapsDisplay::processMessage(
...
+      trajectory.resize(submap_entries.size());
+    }
   }
 }
 
 void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
+  ::cartographer::common::MutexLocker locker(&mutex_);
```
damonkohler: Does this work? The mutex is non-reentrant and QueryOrRenderInProgress() on line 150 will try to take it again.

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -149,6 +162,53 @@ void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
   }
 }
 
+void SubmapsDisplay::AllEnabledChanged() {
+  ::cartographer::common::MutexLocker locker(&mutex_);
+  bool visibility = visibility_all_enabled_->getBool();
```
damonkohler: const

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -149,6 +162,53 @@ void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
...
+  for (auto &trajectory : trajectories_) {
+    for (auto &submap : trajectory) {
+      DisconnectVisibilityChange(submap);
+      submap->SetVisibility(visibility);
+      ConnectVisibilityChange(submap);
+    }
+  }
+}
+
+void SubmapsDisplay::VisibilityChanged(DrawableSubmap* submap) {
```
damonkohler: DrawableSubmap* const submap

cartographer_ros_msgs/msg/SubmapEntry.msg
```diff
@@ -12,5 +12,5 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-uint32 submap_version
+int32 submap_version
```
damonkohler: Why should we change this?

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -149,6 +162,53 @@ void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
...
+  ::cartographer::common::MutexLocker locker(&mutex_);
+  // Handle the case when the user was holding Ctrl,
+  // in which case the neighbouring submaps should also be toggled
+  auto modifiers = QApplication::keyboardModifiers();
+  if (modifiers.testFlag(Qt::ControlModifier)) {
+    bool visibility = submap->GetVisibility();
```
damonkohler: const

cartographer_ros_msgs/msg/SubmapEntry.msg
ojura: Hi, thanks on the response! 

The changes in this pull request have already been organized in separate, self-standing commits (for example, you can find the answer to this question in the description of the commit "Change SubmapEntry submap_version field to be int32"). I imagined that you would go through each of them, in the order they were commited, if you wanted to tackle them one-by-one (please also have a look at the commit descriptions). If it's easier for you, I can still make separate pull requests; just let me know.

For now, I will go address the style nits now and push them as a new commit.

damonkohler:
Thanks. Since we always squash and merge, submitting it as separate PRs allows us to easily rollback if necessary.

cartographer_ros_msgs/msg/SubmapEntry.msg
damonkohler: Thanks. Since we always squash and merge, submitting it as separate PRs allows us to easily rollback if necessary.

cartographer_rviz/cartographer_rviz/submaps_display.cc
ojura: Each DrawableSubmap has its own mutex. Line 130 locks the SubmapsDisplay mutex, while QueryOrRenderInProgress() locks the individual DrawableSubmap mutex. Is this a problem?
I have not observed problems with this, and as far as I have tested, you can't crash anymore it by mashing the submaps plugin enable/disable checkbox.

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -62,19 +65,25 @@ class DrawableSubmap : public QObject {
   bool MaybeFetchTexture(ros::ServiceClient* client);
 
   // Returns whether an RPC is in progress.
```
ojura: Actually, it behaves the same as before -  since the rendering method DrawableSubmaps:UpdateSceneNode() is the one which sets query_in_progress_=false, and it is guarded by a mutex until rendering is finished, this effectively behaved like QueryOrRenderInProgress() even when it was called QueryInProgress. 

My motivation for this was to preserve the original behaviour. I had to split this flag into two separate flags for querying and rendering (it turned out that I need to wait only for query to end in the destructor; rendering never happened after the destructor was called, so I would wait indefinitely).

Will update the comment.

cartographer_rviz/cartographer_rviz/submaps_display.cc
ojura: Are you sure? Since this is a Qt slot, we would need to write that const in all SIGNAL() and SLOT() macros.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -73,11 +80,21 @@ DrawableSubmap::DrawableSubmap(const int trajectory_id, const int submap_index,
   material_->setCullingMode(Ogre::CULL_NONE);
   material_->setDepthBias(-1.f, 0.f);
   material_->setDepthWriteEnabled(false);
-  scene_node_->attachObject(manual_object_);
+  visibility_ = cartographer::common::make_unique<::rviz::BoolProperty>(
```
ojura: https://gist.github.com/ojura/806910b464e76e59915a3955e123298b :-)

ojura:
@damonkohler, I have split this pull request into several minor pull requests. I have also posted a few replies to your nits above here, in this pull request.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
damonkohler: Yes, we're not always perfect :) However, within this file I believe it's always ::cartographer.

cartographer_rviz/cartographer_rviz/submaps_display.cc
damonkohler: In those cases, I don't think it matters since the macros are hiding the implementation. Here, you're never changing the value of the submap argument. So, it can be const. Not a const pointer, but a const arg.

ojura: Okay. I added 'const' everywhere

ojura: Edit: In the end, I added it only to line 177 as you originally said. Qt won't connect a signal with a pointer to non-const to a slot which receives a pointer to const.

PR #293, opened 2017-03-29
Fix race condition

ojura:
See individual commits.

### DIFF
--- a/cartographer_rviz/cartographer_rviz/drawable_submap.cc
+++ b/cartographer_rviz/cartographer_rviz/drawable_submap.cc
@@ -78,6 +78,9 @@ DrawableSubmap::DrawableSubmap(const int trajectory_id, const int submap_index,
 }
 
 DrawableSubmap::~DrawableSubmap() {
+  if (QueryInProgress()) {
+    rpc_request_future_.wait();
+  }
   Ogre::MaterialManager::getSingleton().remove(material_->getHandle());
   if (!texture_.isNull()) {
     Ogre::TextureManager::getSingleton().remove(texture_->getHandle());
--- a/cartographer_rviz/cartographer_rviz/submaps_display.cc
+++ b/cartographer_rviz/cartographer_rviz/submaps_display.cc
@@ -114,13 +114,13 @@ void SubmapsDisplay::processMessage(
 }
 
 void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
+  ::cartographer::common::MutexLocker locker(&mutex_);
   // Update the fading by z distance.
   try {
     const ::geometry_msgs::TransformStamped transform_stamped =
         tf_buffer_.lookupTransform(map_frame_property_->getStdString(),
                                    tracking_frame_property_->getStdString(),
                                    ros::Time(0) /* latest */);
-    ::cartographer::common::MutexLocker locker(&mutex_);
     for (auto& trajectory : trajectories_) {
       for (auto& submap : trajectory) {
         submap->SetAlpha(transform_stamped.transform.translation.z);
@@ -138,7 +138,7 @@ void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
         ++num_ongoing_requests;
       }
     }
-    for (int submap_index = trajectory.size() - 1;
+    for (int submap_index = static_cast<int>(trajectory.size()) - 1;
          submap_index >= 0 &&
          num_ongoing_requests < kMaxOnGoingRequestsPerTrajectory;
          --submap_index) {


cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -159,13 +168,16 @@ void DrawableSubmap::SetAlpha(const double current_tracking_z) {
 
 void DrawableSubmap::UpdateSceneNode() {
   ::cartographer::common::MutexLocker locker(&mutex_);
+  render_in_progress_ = false;
+  if (quitting_) {
```
wohe: Isn't the only benefit of adding the logic for `quitting_` and `render_in_progress_` to save a tiny bit of performance? I'd then argue for keeping the current code. That it finishes the query by updating the texture which is subsequently destroyed seems a tiny bit more work in exchange for simpler code.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -78,6 +78,9 @@ DrawableSubmap::DrawableSubmap(const int trajectory_id, const int submap_index,
 }
 
 DrawableSubmap::~DrawableSubmap() {
+  ::cartographer::common::MutexLocker locker(&mutex_);
+  quitting_ = true;
+  locker.Await([this]() REQUIRES(mutex_) { return !query_in_progress_; });
```
wohe: Seems line 81 and 83 fix the issue and provide the safety mentioned in the commit description. Keep only these?

cartographer_rviz/cartographer_rviz/drawable_submap.cc
ojura: Have a look at my reply to @damonkohler: https://github.com/googlecartographer/cartographer_ros/pull/289#discussion_r108674387

wohe: TL;DR: I think the issue can be fixed by replacing this commit by just adding to the beginning of the destructor:
```
if (query_in_progress_) {
  rpc_request_future_.wait();
}
```
Does this solve the observed issues for you?

You have to be more specific for me to follow your reasoning. For example, you set `quitting_` to true in the destructor, but use it in `MaybeFetchTexture()` line 125. Why would this member function be called if the object was destroyed? It seems this should and does never happen. Line 172 seems to defend against the fact that the object to which `this` points was already destroyed. At this point looking at `quitting_` is too late as the object is already gone.

My investigation: In the problematic case, the main thread in `submaps_display.cc` calls `trajectories_.clear()` while a query is in progress. We need to wait before we destroy the `DrawableSubmap` object. Waiting for `query_in_progress_` to become false is not possible, since it causes a deadlock: The destructor runs in the main thread, so blocks Qt from handling signals. If instead we just wait until the signal is emitted (that's my code suggestion above), Qt knows not to call a slot if it was deleted after emitting as far as I understand.

ojura: Hi! Sorry that I was not specific enough. You are exactly right about everything. I agree, your solution is more elegant; by using the `wait` method of the request future object, there is no need to split the `query_in_progress_` flag. However, it still might be more accurate to rename the existing flag (without splitting it) to "query or render in progress" (or something more specific instead of "render", such as "texture update"), because effectively it really behaves that way.

> You have to be more specific for me to follow your reasoning. For example, you set quitting_ to true in the destructor, but use it in MaybeFetchTexture() line 125. Why would this member function be called if the object was destroyed? It seems this should and does never happen. Line 172 seems to defend against the fact that the object to which this points was already destroyed. At this point looking at quitting_ is too late as the object is already gone.

That is right. Adding the check for `quitting_` was a precautionary (more like desperate :-) ) measure in case we would be extremely unlucky and the destructor is invoked just after fetch completes and emits the `RequestSucceeded()` signal, which would be caught by Qt. The destructor and `UpdateSceneNode()` both compete to lock the mutex; destructor wins and deletes the object. If `UpdateSceneNode()` now reads from a deleted object, the least I could do is read the quitting boolean, which should still be intact, and we know that we should give up. If you're certain that this scenario is impossible, great! We can then remove `quitting_` altogether.

To summarize:
1) Remove `render_in_progress_`, use `rpc_request_future_.wait()`, keep existing `query_in_progress_` as it is, possibly rename it to reference that it also includes updating the textures
2) In `SubmapsDisplay::update()`, I stil think that we should have the scope of the locker for the whole function, because we are also iterating through `trajectories_` in the second part (you weren't suggesting to remove that, only the second commit, correct?)
3) If you're certain that the signal emitted before/during `DrawableSubmap` destruction by the fetcher won't result in a call of `UpdateSceneNode()` where it acquires the mutex after destruction, we can remove `quitting_`.

wohe: Yes, sounds good. I was only commenting on the second commit, the first one can be merged as is.

ojura: One small thing though. Shouldn't reading of query_in_progress_ in the destructor be protected with a mutex? Imagine that destructor reads query_in_progress == false, and immediately after, MaybeFetchTexture() in another thread initializes it.

wohe: I'd say you are overthinking it, and I don't think your suggestion would make it safer.

If it were possible for the outside to call a public member function (e.g. `MaybeFetchTexture()` as you said) and call the destructor before this function call finishes, I don't think you can have correct behavior without keeping some data outside of the object. The destructor might have completed before the called function starts executing its first line in a different thread, and would try to take a mutex that has already been destroyed. In short: we depend on the destructor only being called when there are no outstanding calls to any of the public member functions, and I think this is correct for us to assume.

ojura: I started preparing the updated version of this pull request for you, and just before pushing everything as we agreed on above, I remembered another reason why I introduced the splitting of `query_in_progress_`. 

In the second pull request, where you can toggle individual submap visibility, I added deferring the texture update in `UpdateSceneNode()` for hidden submaps. Only when you toggle their visibility back, the additional call of `UpdateSceneNode()`, which results from clicking the checkbox (and not from a finished fetch thread) will update their texture. This has been added in commit 6f4095b226e1.

I now find setting `query_in_progress_ = false` in `UpdateSceneNode()` a bit awkward, as there could be another query in progress, unrelated to user's clicking of the checkbox. Damon wanted these two pull requests separated, so the splitting made less sense on its own in this pull request.

With that in mind, the options are now the following:
 *  discard commit 6f4095b226e1 in the other pull request, and always fully run `UpdateSceneNode()`, no matter whether the submap is hidden; this pull request is updated as we agreed above
  * we keep the two separate flags, while removing `quitting_`, and keep the deferred texture update in the other pull request

wohe: Let's do the simple fix for the multi-threading crash here to get this merged and discuss splitting `query_in_progress_` in the other PR then.

ojura:
Okay, this should be it.

PR #294, opened 2017-03-29
Add submap toggling

ojura:
See individual commits.

Note: this pull request depends on #293, you should check out that one first.


ojura:
Besides the nits you will probably find, it now remains to be addressed if the optimization in 6f4095b should be applied and how (because it requires splitting of the flag `query_in_progress_`). It's the main culprit which causes the merge conflict at the moment.

Do you think texture updates should be deferred until the maps are set to be visible? In one case, it's a good thing if we are not really interested in the majority of the submaps -  imagine that we turn all maps off at the beginning besides the few interesting ones. It's good that we don't waste resources updating the textures for maps we don't need. However, if we decide to click "All Enabled" at the end, we will have to wait for half a second - a second, because all of the texture updates have been deferred and have to be performed at that single point, which makes the user interface seem a bit laggy in that moment.

ojura:
Hello! To simplify things, I updated the pull request to remove that `query_in_progress_` flag splitting and deferred rendering I was talking about earlier; we can do that in another pull request.

SirVer:
Sorry, this fell through the crackz. I will give this a try this week.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -25,11 +25,16 @@
 #include "Eigen/Geometry"
 #include "OgreGpuProgramParams.h"
 #include "OgreImage.h"
+#include "cartographer/common/make_unique.h"
 #include "cartographer/common/port.h"
 #include "cartographer_ros_msgs/SubmapQuery.h"
 #include "eigen_conversions/eigen_msg.h"
 #include "ros/ros.h"
 
+#if (QT_VERSION < QT_VERSION_CHECK(5, 0, 0))
```
SirVer: This macro looks really scary. Could you change this into a function that is just identity in the else case.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -73,7 +80,12 @@ DrawableSubmap::DrawableSubmap(const int trajectory_id, const int submap_index,
   material_->setCullingMode(Ogre::CULL_NONE);
   material_->setDepthBias(-1.f, 0.f);
   material_->setDepthWriteEnabled(false);
-  scene_node_->attachObject(manual_object_);
+  visibility_ = ::cartographer::common::make_unique<::rviz::BoolProperty>(
+      "", initial_visibility, "", submap_category, SLOT(ChangeVisibility()),
```
SirVer: can you add comments what the empty string means, for example

    "" /* name */, initial_visibility, "" /* category */

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -54,7 +59,9 @@ std::string GetSubmapIdentifier(const int trajectory_id,
 }  // namespace
 
 DrawableSubmap::DrawableSubmap(const int trajectory_id, const int submap_index,
-                               Ogre::SceneManager* const scene_manager)
+                               Ogre::SceneManager* const scene_manager,
+                               ::rviz::Property* submap_category,
+                               const bool initial_visibility)
```
SirVer: s/initial_visibility/show/

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -110,6 +122,18 @@ void DrawableSubmap::Update(
     // for this submap.
     UpdateTransform();
   }
+  visibility_->setName(QStringLiteral("%1-%2.%3")
```
SirVer: Can we get one fold per trajectory? 

If you run multi trajectory in parallel (which does not yet work, but will eventually), it would be better to have the submaps accessible categorized by trajectory.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -110,6 +122,18 @@ void DrawableSubmap::Update(
...
+                           .arg(trajectory_id_)
+                           .arg(submap_index_)
+                           .arg(metadata_version_));
+  visibility_->setDescription(
```
SirVer: the CTRL functionality seems a bit hacky and unnecessary. Please remove.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -244,4 +268,17 @@ float DrawableSubmap::UpdateAlpha(const float target_alpha) {
   return current_alpha_;
 }
 
+void DrawableSubmap::ChangeVisibility() {
```
SirVer: s/Change/Toggle/

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -68,13 +71,22 @@ class DrawableSubmap : public QObject {
   // 'current_tracking_z'.
   void SetAlpha(double current_tracking_z);
 
+  int submap_index() const { return submap_index_; }
+  int trajectory_id() { return trajectory_id_; }
```
SirVer: All these getters can be const() like submap_index

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -95,11 +107,13 @@ class DrawableSubmap : public QObject {
   Eigen::Affine3d slice_pose_ GUARDED_BY(mutex_);
   std::chrono::milliseconds last_query_timestamp_ GUARDED_BY(mutex_);
   bool query_in_progress_ = false GUARDED_BY(mutex_);
+  bool render_in_progress_ = false GUARDED_BY(mutex_);
```
SirVer: seems unused? remove.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -73,7 +80,12 @@ DrawableSubmap::DrawableSubmap(const int trajectory_id, const int submap_index,
   material_->setCullingMode(Ogre::CULL_NONE);
   material_->setDepthBias(-1.f, 0.f);
   material_->setDepthWriteEnabled(false);
-  scene_node_->attachObject(manual_object_);
+  visibility_ = ::cartographer::common::make_unique<::rviz::BoolProperty>(
```
SirVer: no need for a unique_ptr here IIUC. make it a value member in the class and initialize in the initializer list.

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -46,7 +47,9 @@ class DrawableSubmap : public QObject {
   // Each submap is identified by a 'trajectory_id' plus a 'submap_index'.
   // 'scene_manager' is the Ogre scene manager to which to add a node.
   DrawableSubmap(int trajectory_id, int submap_index,
-                 Ogre::SceneManager* scene_manager);
+                 Ogre::SceneManager* scene_manager,
+                 ::rviz::Property* submap_category,
+                 const bool initial_visibility);
```
SirVer: maybe just: visible

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -149,6 +164,51 @@ void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
...
+      DisconnectVisibilityChange(submap);
+      submap->set_visibility(visibility);
+      ConnectVisibilityChange(submap);
+    }
+  }
+}
+
+void SubmapsDisplay::VisibilityChanged(DrawableSubmap* const submap) {
+  ::cartographer::common::MutexLocker locker(&mutex_);
+  // Handle the case when the user was holding Ctrl,
```
SirVer: please remove the CTRL convenience. I think the whole function can then go away.

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -149,6 +164,51 @@ void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
...
+        neighbour_submap->set_visibility(visibility);
+        ConnectVisibilityChange(neighbour_submap);
+      } catch (std::out_of_range& err) {
+      }
+    }
+  }
+}
+
+void SubmapsDisplay::DisconnectVisibilityChange(
+    const std::unique_ptr<DrawableSubmap>& submap) {
```
SirVer: Nit: I would prefer to take the `const DrawableSubmap* const` ptr directly instead of the reference to the unique_ptr.

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -149,6 +164,51 @@ void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
...
+  ::QObject::disconnect(submap.get(),
+                        SIGNAL(VisibilityChanged(DrawableSubmap*)), this,
+                        SLOT(VisibilityChanged(DrawableSubmap*)));
+}
+
+void SubmapsDisplay::ConnectVisibilityChange(
```
SirVer: same here

cartographer_rviz/cartographer_rviz/drawable_submap.cc
ojura: Too bad you don't like it, I really do :-) Since all of the submaps overlap (because laser fans are always inserted in two most recent submaps), toggling one individual submap off at the time is not visually striking - it just makes one part of the map a little darker. By also toggling the neighbouring submaps at the same moment, we get the visual effect that we've really toggled a nice chunk of the map. However, if you want it removed, I can take it out.

Also, I have another idea - instead removing it altogether, this code construct (SubmapsDisplay receives a signal from one Submap and toggles others) can be modified - not to check if Ctrl was pressed and toggle the neighbours, which I admit is a bit hack-ish, but check instead if any other submaps were earlier shift-click selected in the submap list, and to also toggle them. Let me know what you think.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
ojura: I will go with "visible", which you suggested in the header.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
ojura: Sure. Although, this is an official fallback, taken from Qt5 headers.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
ojura: If I remember correctly, I had a good reason to introduce the unique_ptr - because [here](https://github.com/googlecartographer/cartographer_ros/pull/294/commits/cd50b20a5792bdb684c769571dcc54d257a4fe4e#diff-549491b98125fc0df2c02d4d2023a371R126) I purge old submaps in case I detect there's now less of them, which happens when you relaunch the Cartographer node. The Boolproperty objects have the Category object as their Qt parent, and not the DrawableSubmap - so they will survive the destruction, because Qt will not destroy them when we destroy the DrawableSubmap. So it was either adding a delete in the DrawableSubmap destructor, or using a unique pointer, and I went with the latter.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
SirVer: Can you not use that directly? I mean, why do you need to redefine their workaround here?

Could you link where you found it in the code?

cartographer_rviz/cartographer_rviz/drawable_submap.cc
SirVer: That makes sense. Could you add this as a comment?

cartographer_rviz/cartographer_rviz/drawable_submap.cc
ojura: QStringLiteral has been introduced in Qt5 for compile-time string formatting. On indigo, we are building against Qt4. QStringLiteral is not defined there at all. I copied the fallback for compilers which do not support compile-time string formatting from Qt5's version of `qstring.h`.

We could just alias QStringLiteral to QString, but I thought it would be a better idea to [copy their fallback](http://stackoverflow.com/a/21412477/1290505) from `qstring.h`:

```
#ifndef QStringLiteral
// no lambdas, not GCC, or GCC in C++98 mode with 4-byte wchar_t
// fallback, return a temporary QString
// source code is assumed to be encoded in UTF-8

# define QStringLiteral(str) QString::fromUtf8("" str "", sizeof(str) - 1)
#endif
```

cartographer_rviz/cartographer_rviz/drawable_submap.cc
SirVer: You are trying to add a UI concept that rViz (or Qt) does not support natively. It will always be surprising to the user and hard to discover. And it does not add a ton of benefit here. I strongly suggest leaving it out.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
SirVer: Can we use QString instead and format at runtime? I'll take the minimal performance hit in rVize any day over simpler code.

ojura:
Okay, I resolved the style nits and added grouping by trajectories. The hold-ctrl thing has also been removed. (_Sad face._)

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -110,11 +122,20 @@ void DrawableSubmap::Update(
...
+              "Trajectory %1, submap %2, submap version %3")
+          .arg(trajectory_id_)
+          .arg(submap_index_)
+          .arg(metadata_version_));
 }
 
 bool DrawableSubmap::MaybeFetchTexture(ros::ServiceClient* const client) {
   ::cartographer::common::MutexLocker locker(&mutex_);
-  const bool newer_version_available = texture_version_ < metadata_version_;
+  // Received metadata version can also be lower - if we restarted Cartographer
```
SirVer: Nice catch!

ojura:
You're welcome, thank you for making Cartographer so I have something to play with :)

The other thing I had in mind was an optimization which defers texture updates if the submaps are hidden until they are revealed again, as I wrote above (and crossed out).

The plus side of this is that you wouldn't waste time updating the textures you don't need. The minus side is that it requires splitting the `query_in_progress_` flag into `query_in_progress_` and `render_in_progress_`, which increases complexity. Also, if you toggle many hidden submaps visible for the first time at once, you will have a minor lag until their textures are all updated.

The more I think about this, I think it's pointless so I won't open a pull request, but if you want, you can have a look: https://github.com/googlecartographer/cartographer_ros/compare/master...ojura:add-submap-toggling-optim

ojura:
You're welcome, thank you for making Cartographer so I have something to play with :)

The other thing I had in mind was an optimization which defers texture updates if the submaps are hidden, until they are revealed again. I commented above something similar (crossed out), but I figured out another, more elegant way.

You can have a look here: https://github.com/googlecartographer/cartographer_ros/compare/master...ojura:add-submap-toggling-optim

The upside of this is that we won't waste time updating the textures we don't need. The minus side is that it adds a bit of complexity in MaybeFetchFexture(), because we need to release the mutex before calling UpdateSceneNode (emitting the RequestSucceded() Qt signal yields the same result, Qt's meta-object system immediately calls UpdateSceneNode() because the signal is emited from the main thread).

> I'll take the minimal performance hit in rVize any day over simpler code.

I saw that you prefer keeping thing simple, so I am not sure if you'll want to get this merged, because the payoff is not huge, but if you do, let me know.

SirVer:
I think not updating textures that are not visible is not a big gain and comes with headaches - this is basically [cache invalidation](https://martinfowler.com/bliki/TwoHardThings.html). Right now where the default will be most of the time to display them all. I only toggle them on and off for debugging and tuning.

ojura:
Okay. Sometimes I tend to overoptimize things :)

SirVer:
No worries. I can even relate to that :). I think this is the difference between being a maintainer and a contributor. As maintainer I have to think about supporting a feature for all eternity (essentially), while as a contributor I can put all my love into one pet feature and walk away if I choose to do so. That tends to bring different valuing of trade-offs. Thanks for your understanding!

PR #303, opened 2017-04-17
Correct for googlecartographer/cartographer#227.

macmason:
Also includes some clang-format fixups.


macmason:
Although I haven't tested this, I think this PR could land _before_ Cartographer 227 and no builds would break.

SirVer:
@macmason Travis agrees with you. @wohe will be back in the office today.

Issue #308, opened 2017-04-19
separating rviz_plugin from cartographer_ros


jihoonl:
`cartographer_rviz` brings too big dependencies to install in a robot. Could we separate this package out as an independent repo called like `cartographer_ros_desktop`

SirVer:
That is a valid request. This is theoretically simple too, but it is a rather big refactoring that we do not want to tackle right now. We gladly review PRs of course.

jihoonl:
Can you create a new repo for `cartographer_rviz`? Then I will try to create PRs for refactoring

damonkohler:
Done. There's now a cartographer_rviz repo.

SirVer:
I think this just requires an FAQ entry now and can then be closed.

SirVer:
In offline discussions we decided that the options ROS offers for excluding packages from building are sufficient to tackle this. #515 added a FAQ entry to explain how.

PR #315, opened 2017-04-25
cartographer_rviz migration

jihoonl:
As discussed in https://github.com/googlecartographer/cartographer_ros/issues/308

Together with https://github.com/googlecartographer/cartographer_rviz/pull/2


damonkohler:
Thanks for the contribution! As with the cartographer_rviz repo change, you'll need to wait until the Magazino CLA has been processed.

In the meantime, could you please also update the Dockerfiles? Travis will be unhappy until that's complete. If you could migrate the Dockerfile changes to the cartographer_rviz change as well, it would be greatly appreciated!

ojura:
Hello! I have a suggestion. Another way of disabling building cartographer_rviz would be to make your own CMake libcartographer+cartographer_ros superproject, as I have done in https://github.com/ojura/cartographer_superbuild. In the linked superproject's CMakeLists.txt, it would be enough to comment out the line
`add_subdirectory(cartographer_ros/cartographer_rviz)` to avoid building cartographer_rviz.

If your use case is specific enough, maybe this refactoring could be avoided?

ojura:
Another solution would be to use -DCATKIN_BLACKLIST_PACKAGES="cartographer_rviz" when invoking catkin.

jihoonl:
I am just back from vacation. I will try to finish this PR in this week.

SirVer:
Welcome back :)

jihoonl:
I need inputs to make a progress on this PR.

**rviz configuration with demo launche files**
demo launcher files in `cartographer_ros/launch` require `rviz` and `cartographer_rviz` as dependencies. 
It would be more sense to move them out since rviz is separated out from the repo. So, I am thinking about creating `cartographer_demo` package in [cartographer_rviz repo](https://github.com/googlecartographer/cartographer_rviz) that includes all demo launchers. Or should I keep it in `cartographer_ros`?

**cartographer_ros.rosinstall**
[the rosinstall file](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros.rosinstall) is used for [docker](https://github.com/googlecartographer/cartographer_ros/blob/master/Dockerfile.kinetic#L64) and for [installation + demo guide](http://google-cartographer-ros.readthedocs.io/en/latest/). Can we have two seprate rosinstall file? one for docker(without cartographer_rviz) and one for installation(with cartographer_rviz)?

damonkohler:
A `cartographer_demo` package seems like a reasonable idea to me. Please make sure to update the documentation, though.

I'm not sure I understand the concern about the rosinstall file. I assume cartographer_ros's rosinstall file would simply stop referring to cartographer_rviz.

jihoonl:
It is to have a proper rosinstall file that installs `cartographer`, `cartographer_ros`, and `cartographer_rviz` for Sphinx tutorial.

[Building & Installation in cartographer_ros sphinx document](http://google-cartographer-ros.readthedocs.io/en/latest/#building-installation) guides users to use the rosinstall file for installation. Since *Running the demos* section is coming after Building & Installation, many users will try to run demos without `cartographer_rviz` package if I remove `cartographer_rviz` from rosinstall file.

damonkohler:
It seems to me that the documentation for the demo should point to the proposed demo package which can contain the correct rosinstall file. Does that make sense?

damonkohler:
@jihoonl before we take this further, could you explain why building with `-DCATKIN_BLACKLIST_PACKAGES="cartographer_rviz"` is not sufficient for your use case? We've further discussed the ramifications of this change on our end and it will create quite some maintenance burden.

jihoonl:
It will probably work as a workaround. I wanted to have a clear repository separation for robot and for desktop because it is easy to control software in robots. I can manage it internally if you don't this.

damonkohler:
@jihoonl In that case, please investigate what options you have on your side. Hopefully it's easy enough and we can avoid the complications of this PR and having to add yet another repo.

@SirVer found the possibility of sparsely check out only the cartographer_ros package: http://stackoverflow.com/questions/600079/how-do-i-clone-a-subdirectory-only-of-a-git-repository/13738951#13738951

It would be nice to offer some instruction for users who need to do this, so please update the issue #308 with your approach. We can fold it into the FAQ.

Let's keep this PR and the cartographer_rviz repo around until you've confirmed that using build options will work for you.

damonkohler:
@jihoonl any updates? Can we close this PR and delete the rviz repo?

jihoonl:
@damonkohler yes. I will reopen it if it becomes necessary.

ruffsl:
Just a thought, but if we ever intend to release cartographer_ros as installable debian package from a ROS distro repo, it would be nice to have the flexibility to install just the slam aspects of cartographer onto the robot target without roping in all the bulk of desktop dependencies rviz would require. 

Not sure if that's a to-far-off goal, but having packaged version of cartographer_ros would real helpful in enabling other such projects to be packaged and shippable: e.g  cartographer_{turtlebot, fetch, hsr, etc.}.

SirVer:
@ruffsl That is a valid point, but as long as Cartographer hasn't reached 1.0 it seems unwise to invest in that sort of granularity. For now, the suggested installation method is through compilation where the user can pick whatever she likes.

ojura:
@ruffsl Good idea, and don't I think this is in conflict with keeping everything in a single repo - you could have Debian rules scripts which will generate multiple Debian packages for various components of Cartographer, while still keeping everything in a single git repository.

ruffsl:
Oh, I read this issue without looking at the current source structure. My original misconception from this thread was that the `cartographer_ros` package was somehow dependent upon the `cartographer_rviz` codebase. It looks like they are already separate ros packages, and the dependency structure listed from the package.xml is rightfully in the other direction. One can just as easily release separate debian packages for ros packages collected together on a shared repo (it's a common thing for maintainers already). I'm unsure of the real issue here now? Nevermind then.

PR #321, opened 2017-05-02
Added reading intensity values from point clouds

jpapon:
This replaces the reading in assets writer which has hardcoded 1.0 values for intensity.

The downside to this patch is that if a pointcloud2 message without intensity arrives, it will fromROSmsg will throw some warnings.

The fix for that is to add a run time check for the intensity field. I can add that if it is preferred.


SirVer:
Now I feel silly ... I wrote that code. Thanks for digging this up!

Please also sign the CLA.

jpapon:
@SirVer Should be better now.

SirVer:
@jpapon Merge the status as is now. If you disagree with any of the changes, let me know and we can roll back.

PR #322, opened 2017-05-03
Enable multiple configuration for multi-trajectory

yutakaoka:
Hi, I just started to enable multi-trajectory on cartographer_ros. This is some portions for that. Could you review it?


SirVer:
I did not look at the code yet, I think we should get the base branch in. 

But one of the things I wonder immediately is how the user will start a new trajectory with different options. So how do you envision `.lua` files and ROS services to change for this?

yutakaoka:
Yes, I should get [PR of cartographer 248](https://github.com/googlecartographer/cartographer/pull/248) done first.

I am working on making a "start_trajectory" ros service to start a new trajectory with different options and subscribers for new input topics.

SirVer:
Let me know when this is ready for another look.

yutakaoka:
I think this PR could be closed because https://github.com/googlecartographer/cartographer_ros/pull/329 covered this.
Regarding "start_trajectory" service, I will send another PR.

PR #329, opened 2017-05-08
Track Cartographer changes.

SirVer:
Change googlecartographer/cartographer#248 introduced a separation
between MapBuilder and TrajectoryOptions. The ROS repositiories require
changes to track this modification.

Also a slight refactoring in assets_writer.h to only pass explicitly the options required for writing assets and some include what you use fixes.

Tested with demo_backpack_[23]d.launch.

### DIFF
--- a/cartographer_ros/cartographer_ros/assets_writer.cc
+++ b/cartographer_ros/cartographer_ros/assets_writer.cc
@@ -35,12 +35,40 @@ namespace {
 
 namespace carto = ::cartographer;
 
+void WriteTrajectory(const std::vector<::cartographer::mapping::TrajectoryNode>&
+                     trajectory_nodes, const std::string& stem
+                 ) {
+  carto::mapping::proto::Trajectory trajectory;
+  for (const auto& node : trajectory_nodes) {
+    const auto& data = *node.constant_data;
+    auto* node_proto = trajectory.add_node();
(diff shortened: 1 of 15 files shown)


cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -120,8 +135,11 @@ MapBuilderBridge::BuildOccupancyGrid() {
   if (!trajectory_nodes.empty()) {
     occupancy_grid =
         cartographer::common::make_unique<nav_msgs::OccupancyGrid>();
-    cartographer_ros::BuildOccupancyGrid(trajectory_nodes, options_,
-                                         occupancy_grid.get());
+    cartographer_ros::BuildOccupancyGrid(
```
wohe: Add the check that was removed from `cartographer_ros::BuildOccupancyGrid()` here? Rename that function to `BuildOccupancyGrid2D` or so?

cartographer_ros/cartographer_ros/assets_writer.cc
```diff
@@ -35,12 +35,40 @@ namespace {
 
 namespace carto = ::cartographer;
 
+void WriteTrajectory(const std::vector<::cartographer::mapping::TrajectoryNode>&
+                     trajectory_nodes, const std::string& stem
+                 ) {
```
wohe: `clang-format`?

PR #345, opened 2017-05-18
Do not ignore docker history command errors.

ojura:
Original commit by @damonkohler in
googlecartographer/cartographer#298


damonkohler:
Thanks!

PR #358, opened 2017-06-01
Added error handling on start trajectory service

yutakaoka:
I tried to add some error handling functions for start trajectory service for now. If it's not enough, please let me know.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -52,22 +52,23 @@ namespace {
 constexpr int kInfiniteSubscriberQueueSize = 0;
 constexpr int kLatestOnlyPublisherQueueSize = 1;
 
-TrajectoryOptions ToTrajectoryOptions(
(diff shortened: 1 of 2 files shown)


SirVer:
Thanks for adding the error checking! Much appreciated.

I fixed some nits in 03273efe27e463b299c7d96a396ba32dec4de60b. I also renamed `Check*` to something else, because `Check` in google-speak means `crash if false`, i.e. assert.
 In 94547dade1c20450ed3384eb5c87d83553cc5e3f I changed the remaining case I saw where a silent error could slip through. 

@wohe Also pointed out that we do not use ROS logging anywhere else. We rely on the Glog to ROS bridge that is active in the nodes. So for consistency, I changed all ROS logging to using Glog in 8c053594bc812e859ec21a1930656e891927f75e. Sorry for not noticing during the review.

I think this is good to go. Let me know if you are happy with my changes and I'll merge this in.

yutakaoka:
LGTM.
Thank you for making it better. I am always excited about your review.

PR #363, opened 2017-06-08
Add trajectory colors

ojura:
Add color manager for trajectories

Generates colors for different trajectories on request, using a scheme
which ensures even hue distribution, uniform look and high contrast
between consecutive trajectories.

### DIFF
--- a/cartographer_ros/cartographer_ros/color.cc
+++ b/cartographer_ros/cartographer_ros/color.cc
@@ -24,7 +24,7 @@ namespace cartographer_ros {
 
 namespace {
 
-constexpr float kInitialHue = 0.69;
+constexpr float kInitialHue = 0.69f;
 constexpr float kSaturation = 0.85f;
 constexpr float kValue = 0.77f;
 constexpr float kGoldenRatioConjugate = (std::sqrt(5) - 1) / 2.f;
--- a/cartographer_ros/cartographer_ros/color.cc
+++ b/cartographer_ros/cartographer_ros/color.cc
@@ -14,20 +14,21 @@
  * limitations under the License.
  */
 
-#include "cartographer_ros/color_manager.h"
+#include "cartographer_ros/color.h"
 
 #include <cmath>
 
 #include "glog/logging.h"
 
-namespace cartographer_ros {
+namespace {
 
-constexpr float kColorManagerInitialHue = 0.69;
-constexpr float kColorManagerSaturation = 0.85f;
-constexpr float kColorManagerValue = 0.77f;
+constexpr float kInitialHue = 0.69;
+constexpr float kSaturation = 0.85f;
+constexpr float kValue = 0.77f;
 constexpr float kGoldenRatioConjugate = (std::sqrt(5) - 1) / 2.f;
 
-ColorRGB HSVToRGB(const float h, const float s, const float v) {
+::cartographer_ros::ColorRgb HsvToRgb(const float h, const float s,
+                                      const float v) {
   const float h_6 = (h == 1.f) ? 0.f : 6 * h;
   const int h_i = std::floor(h_6);
   const float f = h_6 - h_i;
@@ -53,13 +54,16 @@ ColorRGB HSVToRGB(const float h, const float s, const float v) {
   }
 }
 
-ColorRGB GetColor(int id) {
+}  // namespace
+
+namespace cartographer_ros {
+
+ColorRgb GetColor(int id) {
   CHECK_GE(id, 0);
   // Uniform color sampling using the golden ratio from
   // http://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/
-  const float hue =
-      std::fmod(kColorManagerInitialHue + kGoldenRatioConjugate * id, 1.f);
-  return HSVToRGB(hue, kColorManagerSaturation, kColorManagerValue);
+  const float hue = std::fmod(kInitialHue + kGoldenRatioConjugate * id, 1.f);
+  return HsvToRgb(hue, kSaturation, kValue);
 }
 
 }  // namespace cartographer_ros
--- a/cartographer_ros/cartographer_ros/color.cc
+++ b/cartographer_ros/cartographer_ros/color.cc
@@ -20,6 +20,8 @@
 
 #include "glog/logging.h"
 
+namespace cartographer_ros {
+
 namespace {
 
 constexpr float kInitialHue = 0.69;
@@ -55,8 +57,6 @@ constexpr float kGoldenRatioConjugate = (std::sqrt(5) - 1) / 2.f;
 }
 
 }  // namespace
-
-namespace cartographer_ros {
 
 ColorRgb GetColor(int id) {
   CHECK_GE(id, 0);
--- a/cartographer_ros/cartographer_ros/color_manager.cc
+++ b/cartographer_ros/cartographer_ros/color_manager.cc
@@ -20,22 +20,25 @@
 
 namespace cartographer_ros {
 
-ColorManager::ColorManager(const double initial_hue, const double saturation,
-                           const double value)
-    : current_hue_(initial_hue), saturation_(saturation), value_(value) {}
+constexpr float kColorManagerSaturation = 0.85f;
+constexpr float kColorManagerValue = 0.77f;
+constexpr float kGoldenRatioConjugate = (std::sqrt(5) - 1) / 2.f;
 
-ColorManager::ColorRGB ColorManager::HSVToRGB(const double h, const double s,
-                                              const double v) {
+ColorManager::ColorManager(const float initial_hue)
+    : initial_hue_(initial_hue) {}
+
+ColorManager::ColorRGB ColorManager::HSVToRGB(const float h, const float s,
+                                              const float v) {
   ColorRGB out_rgb;
-  double p, q, t, f, h_6;
+  float p, q, t, f, h_6;
 
-  h_6 = (h == 1.) ? 0. : 6 * h;
+  h_6 = (h == 1.f) ? 0.f : 6 * h;
   int h_i = std::floor(h_6);
   f = h_6 - h_i;
 
-  p = v * (1. - s);
-  q = v * (1. - f * s);
-  t = v * (1. - (1. - f) * s);
+  p = v * (1.f - s);
+  q = v * (1.f - f * s);
+  t = v * (1.f - (1.f - f) * s);
 
   if (h_i == 0)
     out_rgb = {v, t, p};
@@ -50,28 +53,18 @@ ColorManager::ColorRGB ColorManager::HSVToRGB(const double h, const double s,
   else if (h_i == 5)
     out_rgb = {v, p, q};
   else
-    out_rgb = {0., 0., 0.};
+    out_rgb = {0.f, 0.f, 0.f};
 
   return out_rgb;
 }
 
-ColorManager::ColorRGB ColorManager::GetNextColor() {
-  // Uniform color sampling using the golden ratio from
-  // http://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/
-  constexpr double golden_ratio_conjugate = (std::sqrt(5) - 1) / 2.;
-  ColorRGB next_color = HSVToRGB(current_hue_, saturation_, value_);
-  current_hue_ += golden_ratio_conjugate;
-  current_hue_ = std::fmod(current_hue_, 1.);
-  generated_colors_.push_back(next_color);
-  return next_color;
-}
-
 ColorManager::ColorRGB ColorManager::GetColor(int id) {
   CHECK_GE(id, 0);
-  while ((int)generated_colors_.size() <= id) {
-    GetNextColor();
-  }
-  return generated_colors_[id];
+  // Uniform color sampling using the golden ratio from
+  // http://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/
+  const float hue =
+      std::fmod(initial_hue_ + kGoldenRatioConjugate * (id + 8), 1.f);
+  return HSVToRGB(hue, kColorManagerSaturation, kColorManagerValue);
 }
 
 }  // namespace cartographer_ros
--- a/cartographer_ros/cartographer_ros/color_manager.cc
+++ b/cartographer_ros/cartographer_ros/color_manager.cc
@@ -15,7 +15,9 @@
  */
 
 #include "cartographer_ros/color_manager.h"
+
 #include <cmath>
+
 #include "glog/logging.h"
 
 namespace cartographer_ros {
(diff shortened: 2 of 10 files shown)


ojura:
@wohe I have addressed the nits you mentioned in #360. Also, we were discussing the `trajectory_hue_first` option. This was my argument why it could be useful:

https://github.com/googlecartographer/cartographer_ros/pull/360#discussion_r120850521

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -24,12 +24,17 @@
...
+constexpr double kTrajectoryColorManagerValue = 0.75;
 
 MapBuilderBridge::MapBuilderBridge(const NodeOptions& node_options,
                                    tf2_ros::Buffer* const tf_buffer)
     : node_options_(node_options),
       map_builder_(node_options.map_builder_options),
-      tf_buffer_(tf_buffer) {}
+      tf_buffer_(tf_buffer),
+      trajectory_color_manager_(node_options.trajectory_hue_first,
+                                kTrajectoryColorManagerSaturation,
```
wohe: It is best not to eagerly generalize code if it adds complexity. For example it is unclear whether there will ever be a need in this code base for having different saturation and value. So it is best to remove these parameters. If we need them we could still refactor the code then. If we do not, the current code is overly complicated for no good reason.

Similar with the hue option: If there is need, we can introduce it. But it seems doubtful to me at least.

cartographer_ros/cartographer_ros/color_manager.cc
```diff
@@ -0,0 +1,77 @@
...
+  ColorRGB next_color = HSVToRGB(current_hue_, saturation_, value_);
+  current_hue_ += golden_ratio_conjugate;
+  current_hue_ = std::fmod(current_hue_, 1.);
+  generated_colors_.push_back(next_color);
+  return next_color;
+}
+
+ColorManager::ColorRGB ColorManager::GetColor(int id) {
+  CHECK_GE(id, 0);
+  while ((int)generated_colors_.size() <= id) {
```
wohe: No need to keep a list of generated colors. You can directly compute
`hue = std::fmod(kStartHue + kGoldenRatioConjugate * id, 1.f)`

cartographer_ros/cartographer_ros/color_manager.h
```diff
@@ -0,0 +1,52 @@
...
+class ColorManager {
+ public:
+  // A class for online generation of a colour palette, with every two direct
+  //  successors having large contrast. All parameters are from [0, 1].
+  ColorManager(const double initial_hue, const double saturation,
+               const double value);
+
+  struct ColorRGB {
+    // r, g, b are from [0,1]
+    double r;
```
wohe: No need for doubles here since we only use floats.

ojura:
Okay, I have fixed the nits. 

About the configuration parameter for the initial hue of trajectory 0 - if we were to use nav_msgs::Path for the trajectory, we would be able to customize the color in rviz. We opted to use a custom generated line strip marker instead because we want to cover the multi-trajectory scenario, and we sacrificed the ability to customize the color in rviz. However, I believe that we should still provide at least the possibility of changing the initial hue, so that the user can change the color of the trajectory 0 (which should be a common use case) without recompiling Cartographer.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -206,13 +207,19 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
   const auto trajectory_nodes =
       map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
   int marker_id = 0;
-  for (const auto& single_trajectory : trajectory_nodes) {
+  for (int trajectory_id = 0; trajectory_id < (int)trajectory_nodes.size();
```
wohe: Change from `int` to `size_t` to remove the cast? Otherwise, style: C++ casts, i.e. `static_cast<int>(trajectory_nodes.size())`.

cartographer_ros/cartographer_ros/color_manager.cc
```diff
@@ -0,0 +1,70 @@
...
+
+  return out_rgb;
+}
+
+ColorManager::ColorRGB ColorManager::GetColor(int id) {
+  CHECK_GE(id, 0);
+  // Uniform color sampling using the golden ratio from
+  // http://martin.ankerl.com/2009/12/09/how-to-create-random-colors-programmatically/
+  const float hue =
+      std::fmod(initial_hue_ + kGoldenRatioConjugate * (id + 8), 1.f);
```
wohe: Why `+ 8`?

cartographer_ros/cartographer_ros/color_manager.h
```diff
@@ -0,0 +1,47 @@
...
+namespace cartographer_ros {
+
+class ColorManager {
+ public:
+  // A class for online generation of a colour palette, with every two direct
+  //  successors having large contrast. All parameters are from [0, 1].
+  ColorManager(const float initial_hue);
+
+  struct ColorRGB {
+    // r, g, b are from [0,1>
```
wohe: `// r, g, b are form [0, 1].`, right?

cartographer_ros/cartographer_ros/color_manager.h
```diff
@@ -0,0 +1,47 @@
...
+    float r;
+    float g;
+    float b;
+  };
+
+  ColorRGB GetColor(int id);
```
wohe: This doesn't seem to need a class. Make it a free standing function?

cartographer_ros/cartographer_ros/color_manager.cc
```diff
@@ -0,0 +1,70 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "cartographer_ros/color_manager.h"
+#include <cmath>
```
wohe: Style: Empty line to separate related header and C++ headers, empty line to separate C++ headers from the rest.

cartographer_ros/cartographer_ros/color_manager.cc
```diff
@@ -0,0 +1,70 @@
...
+ColorManager::ColorRGB ColorManager::HSVToRGB(const float h, const float s,
+                                              const float v) {
+  ColorRGB out_rgb;
+  float p, q, t, f, h_6;
+
+  h_6 = (h == 1.f) ? 0.f : 6 * h;
+  int h_i = std::floor(h_6);
+  f = h_6 - h_i;
+
+  p = v * (1.f - s);
```
wohe: Style: `const float p = ` here. Similar with the other declarations.

cartographer_ros/cartographer_ros/color_manager.cc
```diff
@@ -0,0 +1,70 @@
...
+  q = v * (1.f - f * s);
+  t = v * (1.f - (1.f - f) * s);
+
+  if (h_i == 0)
+    out_rgb = {v, t, p};
```
wohe: Consistency: Always add curly braces.

Also it seem you could just directly return here and in the other places. No need to first assign to a local variable.

cartographer_ros/cartographer_ros/color_manager.h
ojura: Yes. I mixed it up with the hue.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
ojura: Since trajectory_id is always declared an int in the rest of the same file, I'll use a static cast.

cartographer_ros/cartographer_ros/color_manager.cc
ojura: Whoops. Accidentally commited when I was trying out various colors.

cartographer_ros/cartographer_ros/color_manager.h
ojura: Yes, now that the function is stateless, I can get rid of the whole ColorManager class and just pass the initial hue to GetColor().

ojura:
Further nits have been addressed.

I have a bunch of different .lua config files for various types of robots, sensor configurations, etc. If I can specify in the .lua file a different trajectory colour for each type of robot, when I launch rviz, it's easier to remember which exact type of robot I have launched mapping with. The .lua config file is a convenient place to specify this, if I can't do it already in rviz. If you really don't want this to be an optional parameter in the config file, I'll remove it and have it locally, but it would be easier if I didn't have to rebase it when keeping my fork up to date. :)

wohe:
I see. I still think that adding a lua option is not the right solution here. If we'd have a separate rviz plugin with support for colors that use case would be supported as a side effect. What I fail to see is why the trajectory color is needed here, though: can't you just pick something else to color by robot as a visual cue which is already possible in rviz, e.g. background color, color of laser returns, or so?

ojura:
Yeah, there are other things to choose a colour for. I just liked this one very much :)

This should be it.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -206,13 +206,19 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
   const auto trajectory_nodes =
       map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
   int marker_id = 0;
-  for (const auto& single_trajectory : trajectory_nodes) {
+  for (int trajectory_id = 0;
+       trajectory_id < static_cast<int>(trajectory_nodes.size());
+       trajectory_id++) {
```
wohe: Style nit: `++trajectory_id`

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -206,13 +206,19 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+    const auto& single_trajectory = trajectory_nodes[trajectory_id];
     visualization_msgs::Marker marker;
     marker.id = marker_id++;
     marker.type = visualization_msgs::Marker::LINE_STRIP;
     marker.header.stamp = ::ros::Time::now();
     marker.header.frame_id = node_options_.map_frame;
-    marker.color.b = 1.0;
+    ColorRGB trajectory_color = GetColor(trajectory_id);
```
wohe: nit: `const`

cartographer_ros/cartographer_ros/color_manager.cc
```diff
@@ -0,0 +1,65 @@
...
+#include "glog/logging.h"
+
+namespace cartographer_ros {
+
+constexpr double kColorManagerInitialHue = 0.69;
+constexpr float kColorManagerSaturation = 0.85f;
+constexpr float kColorManagerValue = 0.77f;
+constexpr float kGoldenRatioConjugate = (std::sqrt(5) - 1) / 2.f;
+
+ColorRGB HSVToRGB(const float h, const float s, const float v) {
```
wohe: Move to an unnamed namespace?

cartographer_ros/cartographer_ros/color_manager.cc
```diff
@@ -0,0 +1,65 @@
...
+
+#include "cartographer_ros/color_manager.h"
+
+#include <cmath>
+
+#include "glog/logging.h"
+
+namespace cartographer_ros {
+
+constexpr double kColorManagerInitialHue = 0.69;
```
wohe: `ColorManager` should probably be removed from the name. The file should probably also just be called `color.cc`.

cartographer_ros/cartographer_ros/color_manager.h
```diff
@@ -0,0 +1,37 @@
...
+  float r;
+  float g;
+  float b;
+};
+
+// A function for on-demand generation of a colour palette, with every two
+// direct successors having large contrast. Initial hue is from [0, 1>.
+ColorRGB GetColor(int id);
+
+ColorRGB HSVToRGB(const float h, const float s, const float v);
```
wohe: Style: We prefer to capitalize acronyms as single words, i.e. `HsvToRgb` and `ColorRgb`.

cartographer_ros/cartographer_ros/map_builder_bridge.h
```diff
@@ -23,6 +23,7 @@
 
 #include "cartographer/mapping/map_builder.h"
 #include "cartographer/mapping/proto/trajectory_builder_options.pb.h"
+#include "cartographer_ros/color_manager.h"
```
wohe: No used in the header, so move to the `.cc` file?

cartographer_ros/cartographer_ros/color_manager.cc
```diff
@@ -0,0 +1,65 @@
...
+
+#include "cartographer_ros/color_manager.h"
+
+#include <cmath>
+
+#include "glog/logging.h"
+
+namespace cartographer_ros {
+
+constexpr float kColorManagerInitialHue = 0.69;
```
wohe: These should drop the `ColorManager`. Same with the file: Just `color.cc`?

cartographer_ros/cartographer_ros/color.cc
```diff
@@ -14,20 +14,21 @@
...
 
-#include "cartographer_ros/color_manager.h"
+#include "cartographer_ros/color.h"
 
 #include <cmath>
 
 #include "glog/logging.h"
 
-namespace cartographer_ros {
+namespace {
```
wohe: No need to move this out of the `cartographer_ros` namespace, see e.g. `msg_conversion.cc`.

ojura: Are you sure? E.g. in node.cc you have `kInfiniteSubscriberQueueSize` inside an anonymous namespace, but are using it later inside functions that are inside `::cartographer_ros`.

ojura: Oh wait, sorry, I misunderstood. I will move the anonymous namespace inside cartographer_ros.

ojura:
Travis is currently failing due to something unrelated.

```
/catkin_ws/src/cartographer_ros/cartographer_ros/cartographer_ros/occupancy_grid_test.cc:29:50: error: â€˜ConstantDataâ€™ in â€˜struct cartographer::mapping::TrajectoryNodeâ€™ does not name a type
   const ::cartographer::mapping::TrajectoryNode::ConstantData constant_data{
                                                  ^
/catkin_ws/src/cartographer_ros/cartographer_ros/cartographer_ros/occupancy_grid_test.cc:39:8: error: â€˜constant_dataâ€™ was not declared in this scope
       &constant_data, ::cartographer::transform::Rigid3d::Identity()};
        ^
```

PR #369, opened 2017-06-09
Fix missing ID handling related to starting trajectory

aochiai:
* Remove unused code in `Node::HandleStartTrajectory`.
* Store correct trajectory ID to ROS service response in `Node::HandleStartTrajectory`.
* Register a automatically launched first trajectory to active trajectory list


SirVer:
Thank you very much!

PR #337, opened 2017-05-15
Trajectory and Constraints visualization

jihoonl:
Cleaned up a bit and synced with latest code. It still requires https://github.com/googlecartographer/cartographer/pull/284

PR #375, opened 2017-06-12
Constraints visualization

ojura:
(Sorry, I had to delete my repo and recreate this pull request. Here it is again.)

Originally by @jihoonl. Some nits by @wohe from #337 have been addressed. Some haven't (e.g. https://github.com/googlecartographer/cartographer_ros/pull/337#discussion_r117281636 & https://github.com/googlecartographer/cartographer_ros/pull/337#discussion_r117281982, https://github.com/googlecartographer/cartographer_ros/pull/337#discussion_r117280199). Disclaimer: I have only updated @jihoonl's code so that it compiles with regards to API changes, without giving much thought what it actually does. :)

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -333,18 +333,6 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintList() {
         ToGeometryMsgPoint(constraint_pose.translation()));
     residual_marker->points.push_back(
         ToGeometryMsgPoint(trajectory_node_pose.translation()));
-
-    // Work around the 16384 point limit in RViz by splitting into
-    // multiple markers, similar to GetTrajectoryNodeList().
-    if (constraint_marker->points.size() == 16384) {
-      CHECK_EQ(residual_marker->points.size(), 16384);
-      constraint_list.markers.push_back(*constraint_marker);
-      constraint_marker->id = marker_id++;
-      constraint_marker->points.clear();
-      constraint_list.markers.push_back(*residual_marker);
-      residual_marker->id = marker_id++;
-      residual_marker->points.clear();
-    }
   }
 
   constraint_list.markers.push_back(constraint_intra_marker);
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -24,7 +24,8 @@
 namespace cartographer_ros {
 
 constexpr double kTrajectoryLineStripMarkerScale = 0.07;
-constexpr double kConstraintLineListMarkerScale = 0.025;
+constexpr double kIntraConstraintMarkerScale = 0.025;
+constexpr double kInterConstraintMarkerScale = 0.045;
 
 MapBuilderBridge::MapBuilderBridge(const NodeOptions& node_options,
                                    tf2_ros::Buffer* const tf_buffer)
@@ -253,7 +254,7 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintList() {
   constraint_intra_marker.type = visualization_msgs::Marker::LINE_LIST;
   constraint_intra_marker.header.stamp = ros::Time::now();
   constraint_intra_marker.header.frame_id = node_options_.map_frame;
-  constraint_intra_marker.scale.x = kConstraintLineListMarkerScale;
+  constraint_intra_marker.scale.x = kIntraConstraintMarkerScale;
   constraint_intra_marker.pose.orientation.w = 1.0;
 
   visualization_msgs::Marker residual_intra_marker = constraint_intra_marker;
@@ -267,11 +268,13 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintList() {
   visualization_msgs::Marker constraint_inter_marker = constraint_intra_marker;
   constraint_inter_marker.id = marker_id++;
   constraint_inter_marker.ns = "Inter constraints";
+  constraint_inter_marker.scale.x = kInterConstraintMarkerScale;
   constraint_inter_marker.pose.position.z = 0.1;
 
   visualization_msgs::Marker residual_inter_marker = constraint_intra_marker;
   residual_inter_marker.id = marker_id++;
   residual_inter_marker.ns = "Inter residuals";
+  constraint_inter_marker.scale.x = kInterConstraintMarkerScale;
   residual_inter_marker.pose.position.z = 0.1;
 
   const auto& all_trajectory_nodes =
@@ -288,10 +291,10 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintList() {
       constraint_marker = &constraint_intra_marker;
       residual_marker = &residual_intra_marker;
       // Color mapping for submaps of various trajectories - add trajectory id
-      // to ensure different starting colors. Also add a fixed offset of 20
+      // to ensure different starting colors. Also add a fixed offset of 25
       // to avoid having identical colors as trajectories.
       color_constraint = GetColor(constraint.submap_id.submap_index +
-                                  constraint.submap_id.trajectory_id + 20);
+                                  constraint.submap_id.trajectory_id + 25);
       color_residual.a = 1.0;
       color_residual.r = 1.0;
     } else {
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -245,22 +245,11 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodeList() {
 visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintsList(
     const ::cartographer::mapping::SparsePoseGraph::Constraint::Tag
         wanted_constraint_tag) {
-  visualization_msgs::MarkerArray constraints_viz;
+  visualization_msgs::MarkerArray constraints_list;
 
-  const auto trajectory_nodes =
+  const auto all_trajectory_nodes =
       map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
   const auto constraints = map_builder_.sparse_pose_graph()->constraints();
-  std::vector<cartographer::transform::Rigid3d> submap_transforms;
-  for (auto& single_trajectory : trajectory_nodes) {
-    for (auto& node : single_trajectory) {
-      auto current_trajectory_transforms =
-          map_builder_.sparse_pose_graph()->GetSubmapTransforms(
-              node.constant_data->trajectory_id);
-      std::move(current_trajectory_transforms.begin(),
-                current_trajectory_transforms.end(),
-                std::back_inserter(submap_transforms));
-    }
-  }
 
   int marker_id = 0;
   ros::Time now = ros::Time::now();
@@ -313,10 +302,10 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintsList(
     constraint_marker.points.push_back(submap_pose_point);
     residual_error_marker.points.push_back(submap_pose_point);
     residual_error_marker.points.push_back(trajectory_node_point);
-    constraints_viz.markers.push_back(constraint_marker);
-    constraints_viz.markers.push_back(residual_error_marker);
+    constraints_list.markers.push_back(constraint_marker);
+    constraints_list.markers.push_back(residual_error_marker);
   }
-  return constraints_viz;
+  return constraints_list;
 }
 
 visualization_msgs::Marker MapBuilderBridge::createVisualizationMarker(
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -253,15 +253,13 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintsList(
 
   int marker_id = 0;
   ros::Time now = ros::Time::now();
+  const auto all_submap_data =
+      map_builder_.sparse_pose_graph()->GetAllSubmapData();
   for (const auto& constraint : constraints) {
-    visualization_msgs::Marker constraint_marker, residual_error_marker;
-
-    std_msgs::ColorRGBA color_constraint, color_error;
-
     if (constraint.tag != wanted_constraint_tag) {
       continue;
     }
-
+    std_msgs::ColorRGBA color_constraint, color_error;
     if (constraint.tag ==
         cartographer::mapping::SparsePoseGraph::Constraint::INTER_SUBMAP) {
       // yellow
@@ -276,6 +274,7 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintsList(
       color_error.b = 1.0;
     }
 
+    visualization_msgs::Marker constraint_marker, residual_error_marker;
     constraint_marker = createVisualizationMarker(
         marker_id++, visualization_msgs::Marker::LINE_STRIP, "constraint", now,
         node_options_.map_frame);
@@ -285,23 +284,26 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintsList(
         now, node_options_.map_frame);
     residual_error_marker.color = color_error;
 
-    geometry_msgs::Point submap_point, submap_pose_point, trajectory_node_point;
+    const auto submap_data = all_submap_data[constraint.submap_id.trajectory_id]
+                                            [constraint.submap_id.submap_index];
 
-    submap_point = ToGeometryMsgPoint(
-        submap_transforms[constraint.submap_id.submap_index].translation());
-    submap_pose_point = ToGeometryMsgPoint(
-        (submap_transforms[constraint.submap_id.submap_index] *
-         constraint.pose.zbar_ij)
-            .translation());
-    trajectory_node_point =
-        ToGeometryMsgPoint(trajectory_nodes[constraint.node_id.trajectory_id]
-                                           [constraint.node_id.node_index]
-                                               .pose.translation());
+    const auto submap_origin = submap_data.pose;
+    const auto constraint_pose = submap_origin * constraint.pose.zbar_ij;
+    const auto trajectory_node_pose =
+        all_trajectory_nodes[constraint.node_id.trajectory_id]
+                            [constraint.node_id.node_index]
+                                .pose;
 
-    constraint_marker.points.push_back(submap_point);
-    constraint_marker.points.push_back(submap_pose_point);
-    residual_error_marker.points.push_back(submap_pose_point);
-    residual_error_marker.points.push_back(trajectory_node_point);
+    constraint_marker.points.push_back(
+        ToGeometryMsgPoint(submap_origin.translation()));
+    constraint_marker.points.push_back(
+        ToGeometryMsgPoint(constraint_pose.translation()));
+
+    residual_error_marker.points.push_back(
+        ToGeometryMsgPoint(constraint_pose.translation()));
+    residual_error_marker.points.push_back(
+        ToGeometryMsgPoint(trajectory_node_pose.translation()));
+
     constraints_list.markers.push_back(constraint_marker);
     constraints_list.markers.push_back(residual_error_marker);
   }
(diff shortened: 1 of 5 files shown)


ojura:
This should probably be refactored so it uses a single LINE_LIST marker instead of creating a marker for each constraint, which is very slow. Perhaps, for intra constraints, the new coloring system could be used to color contraints differently for each submap.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -246,6 +247,97 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+         constraint.pose.zbar_ij)
+            .translation());
+    trajectory_node_point =
+        ToGeometryMsgPoint(trajectory_nodes[constraint.node_id.trajectory_id]
+                                           [constraint.node_id.node_index]
+                                               .pose.translation());
+
+    constraint_marker.points.push_back(submap_point);
+    constraint_marker.points.push_back(submap_pose_point);
+    residual_error_marker.points.push_back(submap_pose_point);
```
wohe: I think I understand the `residual_error_marker`, but I'm not sure I understand the purpose of the `constraint_marker`. Could you add a comment about what these markers are showing?

wohe:
Sorry, I did not read your comments first. Yes, I agree.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -246,6 +247,97 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+    std_msgs::ColorRGBA color_constraint, color_error;
+
+    if (constraint.tag != wanted_constraint_tag) {
+      continue;
+    }
+
+    if (constraint.tag ==
+        cartographer::mapping::SparsePoseGraph::Constraint::INTER_SUBMAP) {
+      // yellow
+      color_constraint.a = 1.0;
```
ojura: Sure, that would be a good idea, I'll do it. Do you want to hardcode alpha = 1 in GetColor()?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
ojura: Pinging @jihoonl @domeinzer.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
wohe: Yes. Until we see a use case that needs a different alpha, let's hardcode it.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
domeinzer: The idea behind the constraint marker was to visualize the two points which are connected to each other (especially because we (Magazino) wanted to have constraints on landmarks and have a debugging tool for those constraints.
The residual error marker was just to visualize the remaining error after optimization.

jihoonl: constraint marker visualizes connectivity between trajectory node and submap. It helps to understand which nodes belong to(or connected to) which submap while robot moves.

constraints

In the picture, each number represents origin of a submap(green: local_pose, red:optimized_pose). Yellow line indicates which trajectory node is connected to which submap as inter submap constraints.

wohe:
@ojura Friendly ping. You're not blocked by anyone, right?

ojura:
Hi, sorry, I had some other stuff going on (and spent the bigger part of the last week fixing CMake build errors). I definitely have finishing this on my todo list for this week. If anyone wants to collaborate, feel free to post here or to send me a pull request.

> You're not blocked by anyone, right?
 
I am not sure what you mean :)

wohe:
No problem and no need to hurry. I just wanted to make sure you are not waiting on anyone else to do something before you can continue.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -246,6 +247,97 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+      // cyan
+      color_error.a = 1.0;
+      color_error.b = color_error.g = 1.0;
+    } else {
+      color_constraint.a = color_error.a = 1.0;
+      color_constraint.g = 1.0;
+      color_error.b = 1.0;
+    }
+
+    constraint_marker = createVisualizationMarker(
```
Mofef: I agree on the first part.
But I wouldn't merge the constraint markers and the residual markers into one marker, since then you couldn't disable them individually. Or am I mistaken about this?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
Mofef: You have a point there, @wohe. But maybe this explanation helps: Originally I wanted to see the two end points of a `constraint`. So the `submap` an the `trajectory_node`, because I was adding additional `constraints` and I wanted to see if it works.
Later I wanted to see where the `constraints` are "pulling" a `trajectory_node`. This is what the `residual_error_marker`. 
Since technically it is not only pulling a `trajectory_node` torwards a point on a `submap`, but also a `submap` torwards that `trajectory_node`, it might still be useful to see the `submap` a constraint is attached to (or: whether two `trajectory_nodes` are constrained to the same `submap`.) This is the `constraint_marker`. Now in most use cases the `residual_error_marker` might be the more practical, which is why I would keep them as seperate topics, but with Landmarks in the pipeline this might soon get more relevant again.

jihoonl:
@ojura , @Mofef wanted to create a PR #400 for your fork([larics:constraints_visualization](https://github.com/larics/cartographer_ros/tree/constraints_visualization)). But, somehow, he couldn't create a PR against your repo. Can you apply the changes in #400 in your branch? then we will close #400.

ojura:
Now it's working again same as before. Putting everything inside a single marker and coloring is not yet done.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
ojura: They could be put inside a single marker topic, but inside separate namespaces, similar to how you separated residual and constraint markers.

ojura:
Hello everyone (@wohe @Mofef @jihoonl), please have a look. I completed the refactoring and added coloring. Everything is now published on a single topic, but in separate namespaces, so you can still disable everything.

For intra constraints, I added different coloring for different submaps, while the residuals are now bright red. For inter constraints, I retained Magazino's fixed bright yellow/teal color scheme.

ojura:
By the way, it seems the pixel bleaching code in shaders of the Submaps plugin sometimes clashes with rviz's rendering, which affects the LINE_LIST marker - especially in places where more that 2 submaps are overlaid.

https://www.youtube.com/watch?v=_XMGt5iYzdI

Notice how it affects the grid too.

cartographer_ros/cartographer_ros/node_constants.h
```diff
@@ -33,6 +33,8 @@ constexpr char kSubmapQueryServiceName[] = "submap_query";
 constexpr char kStartTrajectoryServiceName[] = "start_trajectory";
 constexpr char kWriteAssetsServiceName[] = "write_assets";
 constexpr char kTrajectoryNodesListTopic[] = "trajectory_nodes_list";
+constexpr char kConstraintsListTopic[] = "constraints_list";
```
wohe: Use singular? Calling it a list already says it can contain multiple items, i.e. `kConstraintListTopic`, `constraint_list`, etc

cartographer_ros/cartographer_ros/map_builder_bridge.h
```diff
@@ -44,32 +44,33 @@ class MapBuilderBridge {
     TrajectoryOptions trajectory_options;
   };
 
-  MapBuilderBridge(const NodeOptions& node_options, tf2_ros::Buffer* tf_buffer);
+  MapBuilderBridge(const NodeOptions &node_options, tf2_ros::Buffer *tf_buffer);
```
wohe: Please revert to the style used throughout the Cartographer codebase. Did you use `clang-format --style=Google`?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -242,6 +244,113 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+        ToGeometryMsgPoint(constraint_pose.translation()));
+
+    residual_marker->points.push_back(
+        ToGeometryMsgPoint(constraint_pose.translation()));
+    residual_marker->points.push_back(
+        ToGeometryMsgPoint(trajectory_node_pose.translation()));
+
+    // Work around the 16384 point limit in RViz by splitting into
+    // multiple markers, similar to GetTrajectoryNodesList().
+    if (constraint_marker->points.size() == 16384) {
```
wohe: Did `LINE_LIST` also trigger the bug?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -242,6 +244,113 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
   return trajectory_nodes_list;
 }
 
+visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintsList() {
+  visualization_msgs::MarkerArray constraints_list;
+  int marker_id = 0;
+  const ros::Time now = ros::Time::now();
```
wohe: This is used once: inline below?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -242,6 +244,113 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+
+  visualization_msgs::Marker residual_intra_marker = constraint_intra_marker;
+  residual_intra_marker.id = marker_id++;
+  residual_intra_marker.ns = "Intra residuals";
+  residual_intra_marker.pose.position.z = 0.1;
+
+  visualization_msgs::Marker constraint_inter_marker = constraint_intra_marker;
+  constraint_inter_marker.id = marker_id++;
+  constraint_inter_marker.ns = "Inter constraints";
+  constraint_inter_marker.pose.position.z = 0.1;
```
wohe: Why is this different for `constraint_intra_marker`?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -242,6 +244,113 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+
+  const auto all_trajectory_nodes =
+      map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
+  const auto all_submap_data =
+      map_builder_.sparse_pose_graph()->GetAllSubmapData();
+  const auto constraints = map_builder_.sparse_pose_graph()->constraints();
+
+  // Color mapping for submaps of various trajectories - add trajectory id
+  // to ensure different starting colors.
+  const auto submap_color_id_map =
```
wohe: Used once: inline below?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -242,6 +244,113 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+
+  for (const auto& constraint : constraints) {
+    visualization_msgs::Marker *constraint_marker, *residual_marker;
+    std_msgs::ColorRGBA color_constraint, color_residual;
+    if (constraint.tag ==
+        cartographer::mapping::SparsePoseGraph::Constraint::INTRA_SUBMAP) {
+      constraint_marker = &constraint_intra_marker;
+      residual_marker = &residual_intra_marker;
+      color_constraint = GetColor(submap_color_id_map(constraint.submap_id));
+      color_residual.a = 1.0;
```
wohe: Are these colors important, i.e. why not reuse `GetColor()` here as well?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -242,6 +244,113 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+      residual_marker = &residual_inter_marker;
+      // yellow
+      color_constraint.a = 1.0;
+      color_constraint.r = color_constraint.g = 1.0;
+      // cyan
+      color_residual.a = 1.0;
+      color_residual.b = color_residual.g = 1.0;
+    }
+
+    constraint_marker->colors.push_back(color_constraint);
```
wohe: Seems more readable to do this without the temporary variables directly above. If you do not want to repeat the `push_back` twice: loop twice around the `if`?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -242,6 +244,113 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+
+    const auto& submap_data =
+        all_submap_data[constraint.submap_id.trajectory_id]
+                       [constraint.submap_id.submap_index];
+    const auto& submap_origin = submap_data.pose;
+    const auto& trajectory_node_pose =
+        all_trajectory_nodes[constraint.node_id.trajectory_id]
+                            [constraint.node_id.node_index]
+                                .pose;
+    const auto constraint_pose = submap_origin * constraint.pose.zbar_ij;
```
wohe: Let's not use `auto` in these cases. It works now, but I feel it is to easy to miss cases using Eigen types.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -242,6 +244,113 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodesList() {
...
+
+    constraint_marker->colors.push_back(color_constraint);
+    constraint_marker->colors.push_back(color_constraint);
+    residual_marker->colors.push_back(color_residual);
+    residual_marker->colors.push_back(color_residual);
+
+    const auto& submap_data =
+        all_submap_data[constraint.submap_id.trajectory_id]
+                       [constraint.submap_id.submap_index];
+    const auto& submap_origin = submap_data.pose;
```
wohe: Remove or at least rename to `submap_pose`. The origin is `submap_data.pose.translation()`.

cartographer_ros/cartographer_ros/map_builder_bridge.h
ojura: I have, on the whole cartographer_ros codebase (and I have re-run it just now to confirm), and it outputs this. Should I revert these changes nevertheless?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
ojura: Intra constraints are numerous, while the loop closure constraints are rare and should have priority while z-fighting with intra constraints. I will add a comment.

cartographer_ros/cartographer_ros/map_builder_bridge.h
jihoonl: This happened to @Mofef as well. Maybe error in `clang-format`?

wohe: Yes, please. Re-running `clang-format` cannot fix this issue since it changes everything in the file to match the majority style. And right now the majority of cases in this file are clearly different from the rest of the code base.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
ojura: How would you differentiate visually the different types of constraints and markers? Since the intra constraints are the most numerous, I decided to use the color scheme for their markers; for other types of markers, I used fixed colors (e.g. so you know that loop closure constraints are bright yellow).

cartographer_ros/cartographer_ros/map_builder_bridge.cc
ojura: Will test.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
wohe: I'm not opposed to using fixed colors for these, I just can't see how these specific colors are better than say `GetColor(1)`, `GetColor(2)`, and `GetColor(3)`.

cartographer_ros/cartographer_ros/node_constants.h
ojura: Sure. Should I fix the line above too?

cartographer_ros/cartographer_ros/map_builder_bridge.h
ojura: Ah, I see. But it's odd that it decided to this - this style was never the majority.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
ojura: Ah, I see. Currently, we can't generate 100% bright yellow and teal with GetColor(), as the HSV value and saturation are hard coded in colors.cc (as you made me do it :-) ) to produce a bit more muted colors. I would prefer to have loop constraints as 100% bright colors, so they are more easily spotted and differentiated from the rest.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
ojura: In that case I would be assigning the pointer twice, which is very ugly to me. Is it okay if I put a for loop in each branch?

ojura: Also, I cannot get rid of the temporary variables. ColorRGBA does not have a nice constructor that I could easily pass to push_back. (And I do want to retain the special bright colors for the loop closure constraints).

ojura:
I believe it's ready for merging, I have mostly addressed the nits (and the rest is a compromise :-). Please merge #402 first.

ojura:
@wohe, here's some art :-)
http://free.iiifhosting.com/iiif/8cf6f3eea5aa729a5551ab535c75fce49e6a7d71fde965e4f43550d89efe8741/

cartographer_ros/cartographer_ros/map_builder_bridge.cc
ojura: You were right, this wasn't neccessary. Removed.

ojura:
I have two minor questions for the reviewer:
- is it okay to use `const auto&` with e.g.  `GetAllSubmapData()` and the rest
- should I add [here](https://github.com/googlecartographer/cartographer_ros/pull/375/files#diff-aaf9b1c28824a784d19a7190cdf31e8eR320) multiplication with `node.constant_data->tracking_to_pose).translation()` like [here](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/map_builder_bridge.cc#L227)?

ojura:
I have just rebased on top of merged #402. Also, if you're having z-fighting-alike rendering bugs like in the video I linked above, I found that recompiling rviz with the change I described in [this comment](https://github.com/ros-visualization/rviz/issues/1067#issuecomment-312518247) helped.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -242,6 +244,101 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodeList() {
...
+  constraint_inter_marker.id = marker_id++;
+  constraint_inter_marker.ns = "Inter constraints";
+  constraint_inter_marker.pose.position.z = 0.1;
+
+  visualization_msgs::Marker residual_inter_marker = constraint_intra_marker;
+  residual_inter_marker.id = marker_id++;
+  residual_inter_marker.ns = "Inter residuals";
+  residual_inter_marker.pose.position.z = 0.1;
+
+  const auto& all_trajectory_nodes =
```
wohe: `const auto` without the reference for readability since these are returned by value? Same for the submaps and constraints.

wohe:
Regarding the other question: Since we now only publish the translational part it does not matter. The transform I suggested adding is between yaw-only xy plane aligned frame and the full orientation. If you leave it out, the orientation will not be correct. In case you want to add it: it is needed for both the constraint pose and the trajectory node pose. You could also just add a comment why it is not needed here.

ojura:
In that case, would it make sense to remove it from `GetTrajectoryNodeList()` as well, because we are taking only the translational part there too? That would make things consistent. Plus, I can add a comment why it's left out.

wohe:
Sounds good.

ojura:
Done.

wohe:
Thanks, I'll merge as soon as the tests passed.

PR #376, opened 2017-06-14
Adds arg to set cartographer version for docker build.

damonkohler:


ojura:
Shouldn't this be mirrored in Dockerfile.kinetic?

PR #377, opened 2017-06-14
Return ColorRGBA directly from GetColor

ojura:
@wohe here's that small change you mentioned in #375, make `GetColor` return `std_msgs::ColorRGBA` directly.

BTW I have two ways of making this change. In the first version (this pull request), I kept the `ColorRgb` struct for use in `HsvToRgb` because the initializer list (`{r, g, b}`) syntax makes our life easier (`ColorRGBA` can't be constructed directly like this). Another option is in https://github.com/googlecartographer/cartographer_ros/compare/master...larics:color-rgba-2, where I wrote a factory function for `ColorRGBA` which is then used in `HsvToRgb`, and this still looks relatively okay. Let me know which one you prefer. (Or if you have something completely different in mind :) )

### DIFF
--- a/cartographer_ros/cartographer_ros/color.h
+++ b/cartographer_ros/cartographer_ros/color.h
@@ -21,8 +21,8 @@
 
 namespace cartographer_ros {
 
-// A function for on-demand generation of a colour palette, with every two
-// direct successors having large contrast. Initial hue is from [0, 1>.
+// A function for on-demand generation of a color palette, with every two
+// direct successors having large contrast.
 ::std_msgs::ColorRGBA GetColor(int id);
 
 }  // namespace cartographer_ros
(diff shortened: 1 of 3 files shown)


wohe:
Seems not a huge difference, but I'd say the alternative with the factory function is slightly more readable.

ojura:
Sure. I've updated the pull request to use that version.

cartographer_ros/cartographer_ros/color.h
```diff
@@ -17,18 +17,13 @@
...
-struct ColorRgb {
-  // r, g, b are from [0,1]
-  float r;
-  float g;
-  float b;
-};
+namespace cartographer_ros {
 
 // A function for on-demand generation of a colour palette, with every two
 // direct successors having large contrast. Initial hue is from [0, 1>.
```
wohe: Remove the sentence about the initial hue?

cartographer_ros/cartographer_ros/color.h
```diff
@@ -17,18 +17,13 @@
...
 
-struct ColorRgb {
-  // r, g, b are from [0,1]
-  float r;
-  float g;
-  float b;
-};
+namespace cartographer_ros {
 
 // A function for on-demand generation of a colour palette, with every two
```
wohe: Consistency: US English, so `colour` -> `color`.

Issue #368, opened 2017-06-08
Keep spinning while doing the final optimization in the offline node


ojura:
This way the effects of the final optimization can also be viewed live in rviz, since everything will still be published (now it's not, because spinning is done only in the bag message processing loop).

Ideas: launch an asyncspinner to handle this. If it's problematic that the time stopped after the bag is finished (e.g. rviz won't redraw the transforms if the time is not moving), the simulation clock could be faked to continue at the same rate as the wall clock when the bag ends.

Also, it could be useful to add an option to keep the offline node running after it saves the assets. Perhaps we still want to look around the map in rviz, and for that we need to keep publishing the submap list.

SirVer:
I agree that this would be nice, but your suggested solutions sound quite hackish. I think we  have to acknowledge that ROS does not allow what we'd like here - basically a distinction between data-time and real-time and we have to decide if we are willing to work around or not.

> Also, it could be useful to add an option to keep the offline node running after it saves the assets. Perhaps we still want to look around the map in rviz, and for that we need to keep publishing the submap list.

This I do not like. The quality of the final map should be judged by inspecting the assets, i.e. x-rays and other outputs. The rViz representation is just an approximation really. 

Also, the node can exit, as long as rViz keeps running, it should still display the data, or not? We could change the launch files to remove the `required` keyword on the node, but I do not want to do that because it seems to be the reasonable default for the use case.

ojura:
> I agree that this would be nice, but your suggested solutions sound quite hackish.

I know it's hackish, I'm just throwing ideas around :)

> Also, the node can exit, as long as rViz keeps running, it should still display the data, or not? 

That's true, it's still displaying the data but it's easy to disrupt this state. I.e. toggle off/on the visibility of the submaps plugin, or the scan matched point cloud, or the trajectory. These won't get redrawn once they are toggled back on, because they aren't published, since the offline node has quit. If you have a workflow which requires that the map persists in rviz, you're out of luck.

Compare this with the regular online node plus playing a bag - once the bag ends, the node is still alive and publishing everything, listening to services, etc., and you don't have this problem. So I would call it a deficiency of the offline node compared to the offline node.

SirVer:
An alternative would be to replay a bag into the online system at 50x times the speed or so. That is similar to the offline system then, but it stays alive. 

You convinced me though that having a `--do-not-exit` or `--keep-running` option in the online node would help.

PR #380, opened 2017-06-15
Publish more accurate clock in offline node

ojura:
Clock of the message in the msg variable is one second later than the
message in delayed_message. It should be more correct to publish the
clock of the message which really has been processed in SLAM and which
affects the position and the map, rather than the one that has only
been pushed into the delayed_msgs queue.

Also change the status message printing not to include the one second
buffer into the amount of bag processed seconds. Spinning the ROS
event queue has also been moved into the inner loop.

### DIFF
--- a/cartographer_ros/cartographer_ros/offline_node_main.cc
+++ b/cartographer_ros/cartographer_ros/offline_node_main.cc
@@ -237,6 +237,16 @@ void Run(const std::vector<string>& bag_filenames) {
               ->HandleOdometryMessage(
                   topic, delayed_msg.instantiate<nav_msgs::Odometry>());
         }
+        rosgraph_msgs::Clock clock;
+        clock.clock = delayed_msg.getTime();
+        clock_publisher.publish(clock);
+
+        ::ros::spinOnce();
+
+        LOG_EVERY_N(INFO, 100000)
+            << "Processed " << (delayed_msg.getTime() - begin_time).toSec() << " of "
+            << duration_in_seconds << " bag time seconds...";
+
         delayed_messages.pop_front();
       }
 
@@ -247,15 +257,6 @@ void Run(const std::vector<string>& bag_filenames) {
       }
       delayed_messages.push_back(msg);
 
-      rosgraph_msgs::Clock clock;
-      clock.clock = msg.getTime();
-      clock_publisher.publish(clock);
-
-      ::ros::spinOnce();
-
-      LOG_EVERY_N(INFO, 100000)
-          << "Processed " << (msg.getTime() - begin_time).toSec() << " of "
-          << duration_in_seconds << " bag time seconds...";
     }
 
     bag.close();


SirVer:
Mmh, the problem is only in everything that uses ros::Time::now(), which is `GetSubmapList` and `GetTrajectoryNodesList`. Right now they claim use the current time, which is the TF time, which is actually 1sec ahead of the data.

Your suggested change publishes TF messages in the future, i.e. the clock is consistent with all other messages and Cartographer's published data is on the same clock as its data. I guess this is an improvement.

Fundamentally, Cartographer should be data driven, i.e. should always use the latest timestamp of its data to actually publish. Unfortunately, this does not mesh well with ROS's design - for example it becomes impossible to continue publishing in the offline node when no more data is coming in.

SirVer:
gotta run for today, will merge this on Monday.

ojura:
> Mmh, the problem is only in everything that uses ros::Time::now(), which is GetSubmapList and GetTrajectoryNodesList.

...and also all other nodes which might potentially be subscribed to the offline node and which use ros::Time::now(). Also, there is a case in `PublishTrajectoryStates` where the published transform is assigned the ros::Time::now() stamp, which is wrong.

> Your suggested change publishes TF messages in the future

I am not sure I understood this, in fact, I think you're mistaken here (but everything else you wrote is correct). This change only moves clock back one second into the past (as in the `msg` message), so that it's consistent with everything else, as you have correctly said. Nothing is published into the future, only the TF buffer is given a glimpse into the future. Previously (before #342), the TF buffer had the whole bag's worth of transforms loaded inside. But you wouldn't say that we were publishing TF message in the future.

In whole cartographer, the _only_ places where these messages from the future (<1s older than the head of the queue) have any impact at all are the `delayed_msgs` queue itself, and the TF buffer. The old behaviour (before #342) was to publish the clock of the message which had just been dispatched to the sensor bridge. This restores that behaviour.

ojura:
> Unfortunately, this does not mesh well with ROS's design - for example it becomes impossible to continue publishing in the offline node when no more data is coming in.

Yes, we discussed this is #368. In the end, in my use case, I had to implement the hack I proposed there (publishing a fake clock when the bag ends). Many things in ROS use the simulation time when they should use wall time (e.g. registering/waiting for a new service, tf view_frames, publishing interactive markers...) and become completely broken if the simulation time comes to a standstill, like at the end of the bag. Hopefully they will change this in ROS 2.

SirVer:
> I am not sure I understood this, in fact, I think you're mistaken here (but everything else you wrote is correct).

Well, [we do publish] (https://github.com/larics/cartographer_ros/blob/c7ce9cd9d65683a84958373a28c2dbdc053b2746/cartographer_ros/cartographer_ros/offline_node_main.cc#L191) the TF messages. So we are publishing them in the future. Not user if any node will care about this.

Anyways, I think this is an improvement. Gonna merge! Thanks for the contribution again!

ojura:
Oh, sorry, totally forgot about that! Maybe that could be also moved into the inner loop, so that the `delayed_msg` is published instead.

SirVer:
It seems like no solution is 'correct', so lets leave it like this until we find a better way.

ojura:
I agree, after thinking about it a little more - these messages will just get inserted into clients' tf buffers, just like they are inserted in the offline node's tf buffer on purpose. Shouldn't hurt anything.

SirVer:
> these messages will just get inserted into clients' tf buffers, 

That will be the most common use case - but probably not the only. I think it likely that our change will break some things, but as mentioned, we cannot avoid this as long as we stay in ROS.

A proper solution would be to build our own tf_buffer that can deal with hours of data without breaking down computationally or fixing tf2 and reverting to preloading the TFs.

TSC21:
>  fixing tf2 and reverting to preloading the TFs.

Thumbs up for this solution!

ojura:
> build our own tf_buffer

By the way, what about `TransformInterpolationBuffer` used in `assets_writer`?  Why hasn't it been used in the offline or online node?

SirVer:
It only contains one interpolated transform, but for the nodes we require to understand the full TF tree to convert between all the `frame_id`s. We used to have code for this before open sourcing, but we decided to use TF in OSS.

PR #390, opened 2017-06-21
Adds a tool to start a new trajectory from an existing Lua file.

SirVer:


cartographer_ros/cartographer_ros/trajectory_options.cc
```diff
@@ -50,4 +50,40 @@ TrajectoryOptions CreateTrajectoryOptions(
 
   return options;
 }
+
+// Try to convert 'msg' into 'options'. Returns false on failure.
```
wohe: Remove. This should only be in the header.

cartographer_ros/launch/demo_backpack_2d.launch
```diff
@@ -21,6 +21,4 @@
 
   <node name="rviz" pkg="rviz" type="rviz" required="true"
       args="-d $(find cartographer_ros)/configuration_files/demo_2d.rviz" />
-  <node name="playbag" pkg="rosbag" type="play"
```
wohe: Revert?

Issue #392, opened 2017-06-23
libcartographer_ros can't be built as a shared library


ojura:
It fails on linking:

```
[cartographer_ros:make] [ 72%] Linking CXX shared library /home/juraj/catkin_ws/devel/.private/cartographer_ros/lib/libcartographer_ros.so
[cartographer_ros:make] /usr/bin/ld: /home/juraj/catkin_ws/devel/lib/libcartographer.a(lua_parameter_dictionary.cc.o): relocation R_X86_64_PC32 against symbol `_ZNSt6vectorINSt7__cxx1112basic_stringIcSt11char_traitsIcESaIcEEESaIS5_EED1Ev' can not be used when making a shared object; recompile with -fPIC
[cartographer_ros:make] /usr/bin/ld: final link failed: Bad value                                                                    
[cartographer_ros:make] collect2: error: ld returned 1 exit status                                                                   
[cartographer_ros:make] CMakeFiles/cartographer_ros.dir/build.make:803: recipe for target '/home/juraj/catkin_ws/devel/.private/cartographer_ros/lib/libcartographer_ros.so' failed
```

By default, Catkin builds libraries as shared objects. However, libcartographer_ros is built as a static library on Ubuntu <=16.04 completely by accident - when GMock is find_package()-d by `google_enable_testing()`, the [`BUILD_SHARED_LIBS` CMake variable is set to OFF](https://github.com/google/googlemock/blob/master/googlemock/CMakeLists.txt#L10). This doesn't happen on Ubuntu 17.04 (GMock is not built from source, but rather used directly), and so `BUILD_SHARED_LIBS` is[ set to ON by Catkin](https://github.com/ros/catkin/blob/kinetic-devel/cmake/tools/libraries.cmake) as it was previously undefined. I am going to submit a pull request which explicitly sets it to OFF for cartographer_ros to make everything consistent with earlier Ubuntu releases.

However, this way, I found out that building libcartographer_ros as a shared library fails. Perhaps you may consider this to be an issue which needs addressing.

Edit: added link to GMock CMakeLists.

ojura:
Resolution in #392: won't fix, because shared build will not be supported.

PR #393, opened 2017-06-23
Build cartographer_ros as a static library

ojura:
This was implicitly set on <= Ubuntu 16.04 when find_package()-ing
GMock in google_enable_testing(). On 17.04, this doesn't happen, and
Catkin builds a shared library by default. cartographer_ros currently
can't be built as a shared library due to a linking problem (which is
a separate issue - #392). This makes building on 17.04 consistent with
older versions of Ubuntu.

### DIFF
--- a/cartographer_ros/CMakeLists.txt
+++ b/cartographer_ros/CMakeLists.txt
@@ -37,7 +37,7 @@ set(PACKAGE_DEPENDENCIES
 
 find_package(cartographer REQUIRED)
 include("${CARTOGRAPHER_CMAKE_DIR}/functions.cmake")
-option(BUILD_SHARED_LIBS "Build libcartographer_ros as a shared library." OFF)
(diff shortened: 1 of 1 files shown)


SirVer:
I cannot think of a reason why libcartographer_ros should ever be build as a shared library. Can we make this non-optional and always set to static?

ojura:
[CMake manual](https://cmake.org/cmake/help/v3.0/variable/BUILD_SHARED_LIBS.html) says that it's common to set this using `option`. However, if you want, I can change it to be a regular CMake variable defined using `set`.

SirVer:
Yes, go for set I'd say. Then you can also close #392, because it is no longer a bug, but a feature :)

ojura:
Done.

SirVer:
Thank you! You are quickly becoming our most proficient build system specialist...

Issue #343, opened 2017-05-17
Support multi trajectory


SirVer:
googlecartographer/cartographer_ros/#335 started making the node multi trajectory aware. This leaks some problematic abstractions:

1) When a trajectory is finished, the node writes some derived assets (Trajectory + OccupancyGrid or X-Rays + PLY) and starts a new one. The assets use `trajectory_builder_options` for laser insertion options and resolution (which one should this be for multi trajectory?). The trajectory proto only contains the current opinion of one trajectory.
2) BuildOccupancyGrid2D - which is used to publish a legacy OccupancyGrid for ROS.

I think we have to rethink what we want to do here. I think writing the assets is fundamentally wrong in the node - instead, it should only dump the sparse pose graph (with or without submaps, as desired) and all further asset generation should be deferred to assets_writer_main. This would allow us to hold much less data in memory while slamming. 

This would also mean to pull OccupancyGrid generation into a separate ROS node that translates between Cartographer's representation and legacy ROS Occupancy grids. I am unclear if this node should keep laser data around or just reuse the submaps from cartographer and draw them on-top of each other (similar to what the visualization currently does).

/cc @macmason @wohe @yutakaoka Thoughts?

wohe:
Sounds good. We might also want to separate finishing a trajectory from writing the current SLAM state. Maybe we should move this to a separate service?

yutakaoka:
For now(https://github.com/googlecartographer/cartographer_ros/pull/346), I will create `write_assets` service to separate it from `finish_trajectory` service.

PR #394, opened 2017-06-23
Make WriteAssets multi trajectory aware.

SirVer:
Related to #343.

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -89,10 +89,8 @@ void MapBuilderBridge::WriteAssets(const string& stem) {
     }
 
     if (node_options_.map_builder_options.use_trajectory_builder_3d()) {
-      // TODO(yutakaoka): Add multi-trajectory support.
-      CHECK_EQ(all_trajectory_nodes.size(), 1);
       Write3DAssets(
-          all_trajectory_nodes[0],
+          all_trajectory_nodes,
           trajectory_options_[0]
               .trajectory_builder_options.trajectory_builder_3d_options()
               .submaps_options()
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -38,6 +38,7 @@ int MapBuilderBridge::AddTrajectory(
       expected_sensor_ids, trajectory_options.trajectory_builder_options);
   LOG(INFO) << "Added trajectory with ID '" << trajectory_id << "'.";
 
+  // Make sure there is no trajectory with 'trajectory_id' yet.
   CHECK_EQ(sensor_bridges_.count(trajectory_id), 0);
   sensor_bridges_[trajectory_id] =
       cartographer::common::make_unique<SensorBridge>(
@@ -53,6 +54,7 @@ int MapBuilderBridge::AddTrajectory(
 void MapBuilderBridge::FinishTrajectory(const int trajectory_id) {
   LOG(INFO) << "Finishing trajectory with ID '" << trajectory_id << "'...";
 
+  // Make sure there is a trajectory with 'trajectory_id'.
   CHECK_EQ(sensor_bridges_.count(trajectory_id), 1);
   map_builder_.FinishTrajectory(trajectory_id);
   map_builder_.sparse_pose_graph()->RunFinalOptimization();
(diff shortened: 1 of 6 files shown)


cartographer_ros/cartographer_ros/occupancy_grid.cc
```diff
@@ -58,31 +60,34 @@ Eigen::AlignedBox2f ComputeMapBoundingBox(
...
 void BuildOccupancyGrid2D(
-    const std::vector<::cartographer::mapping::TrajectoryNode>&
-        trajectory_nodes,
+    const std::vector<std::vector<::cartographer::mapping::TrajectoryNode>>&
+        all_trajectory_nodes,
     const string& map_frame,
     const ::cartographer::mapping_2d::proto::SubmapsOptions& submaps_options,
     ::nav_msgs::OccupancyGrid* const occupancy_grid) {
-  CHECK(!trajectory_nodes.empty());
+  CHECK(!all_trajectory_nodes.empty());
```
wohe: This now checks that there is a trajectory but we wanted to check if there is some node to compute proper map limits. Also, nodes can be trimmed now so that check would also no longer make sense. If there is no non-trimmed node, we will not have a proper time set. `CHECK` that latest_time is not `min()` anymore afterwards?

cartographer_ros/cartographer_ros/occupancy_grid.cc
```diff
@@ -58,31 +60,34 @@ Eigen::AlignedBox2f ComputeMapBoundingBox(
...
-      ComputeMapLimits(submaps_options.resolution(), trajectory_nodes);
+      ComputeMapLimits(submaps_options.resolution(), all_trajectory_nodes);
   carto::mapping_2d::ProbabilityGrid probability_grid(map_limits);
   carto::mapping_2d::RangeDataInserter range_data_inserter(
       submaps_options.range_data_inserter_options());
-  for (const auto& node : trajectory_nodes) {
-    if (node.trimmed()) {
-      continue;
+  carto::common::Time latest_time =
+      carto::common::FromUniversal(std::numeric_limits<int64_t>::min());
```
wohe: carto::common::Time::min()?

cartographer_ros/cartographer_ros/occupancy_grid.cc
```diff
@@ -58,31 +60,34 @@ Eigen::AlignedBox2f ComputeMapBoundingBox(
...
+  for (const auto& trajectory_nodes : all_trajectory_nodes) {
+    for (const auto& node : trajectory_nodes) {
+      if (node.trimmed()) {
+        continue;
+      }
+      latest_time = std::max(latest_time, node.time());
+      CHECK(node.constant_data->range_data_3d.returns.empty());  // No 3D yet.
```
wohe: Should this be a `TODO` or should be remove the support for 3D from `ComputeMapBoundingBox()`?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -153,20 +153,15 @@ std::unique_ptr<nav_msgs::OccupancyGrid>
...
-       map_builder_.sparse_pose_graph()->GetTrajectoryNodes()) {
-    flat_trajectory_nodes.insert(flat_trajectory_nodes.end(),
-                                 single_trajectory_nodes.begin(),
-                                 single_trajectory_nodes.end());
-  }
+  const auto all_trajectory_nodes =
+      map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
   std::unique_ptr<nav_msgs::OccupancyGrid> occupancy_grid;
-  if (!flat_trajectory_nodes.empty()) {
+  if (!all_trajectory_nodes.empty()) {
```
wohe: This should verify that there is any non-trimmed node, not that there is a trajectory?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -72,16 +72,16 @@ void MapBuilderBridge::SerializeState(const std::string& stem) {
 void MapBuilderBridge::WriteAssets(const string& stem) {
   const auto all_trajectory_nodes =
       map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
-  // TODO(yutakaoka): Add multi-trajectory support.
   CHECK_EQ(trajectory_options_.count(0), 1);
-  CHECK_EQ(all_trajectory_nodes.size(), 1);
   if (all_trajectory_nodes[0].empty()) {
```
wohe: This should look for the existence of at least one non-trimmed node?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -72,16 +72,16 @@ void MapBuilderBridge::SerializeState(const std::string& stem) {
 void MapBuilderBridge::WriteAssets(const string& stem) {
   const auto all_trajectory_nodes =
       map_builder_.sparse_pose_graph()->GetTrajectoryNodes();
-  // TODO(yutakaoka): Add multi-trajectory support.
   CHECK_EQ(trajectory_options_.count(0), 1);
```
wohe: Add a comment that this only checks that trajectory 0 options exist which we will use here? Seems easy to misunderstand.

PR #395, opened 2017-06-23
Support multiple trajectories in the asset_writer_main

### LINKED ISSUE #343: Support multi trajectory

SirVer:
- Also track googlecartographer/cartographer#360.

Related to #343.

### DIFF
--- a/cartographer_ros/cartographer_ros/assets_writer_main.cc
+++ b/cartographer_ros/cartographer_ros/assets_writer_main.cc
@@ -154,7 +154,7 @@ void Run(const string& pose_graph_filename,
   CHECK_EQ(pose_graph_proto.trajectory_size(), bag_filenames.size())
       << "Pose graphs contains " << pose_graph_proto.trajectory_size()
       << " trajectories while " << bag_filenames.size()
-      << " bags where provided. This tool requires one bag for each "
+      << " bags were provided. This tool requires one bag for each "
          "trajectory in the same order as the correponding trajectories in the "
          "pose graph proto.";
 
--- a/cartographer_ros/cartographer_ros/assets_writer_main.cc
+++ b/cartographer_ros/cartographer_ros/assets_writer_main.cc
@@ -136,7 +136,8 @@ std::unique_ptr<carto::io::PointsBatch> HandleMessage(
   return points_batch;
 }
 
-void Run(const string& pose_graph_filename, const std::vector<string>& bag_filenames,
+void Run(const string& pose_graph_filename,
+         const std::vector<string>& bag_filenames,
          const string& configuration_directory,
          const string& configuration_basename, const string& urdf_filename) {
   auto file_resolver =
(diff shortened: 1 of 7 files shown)


SirVer:
Depends on #394 being merged first.

SirVer:
Rebased and ready for review.

cartographer_ros/cartographer_ros/split_string.cc
```diff
@@ -0,0 +1,34 @@
...
+
+#include "cartographer_ros/split_string.h"
+
+#include <sstream>
+
+namespace cartographer_ros {
+
+std::vector<std::string> SplitString(const std::string& input,
+                                     const char delimiter) {
+  std::stringstream stream(input);
```
wohe: `std::istringstream`?

cartographer_ros/cartographer_ros/assets_writer_main.cc
```diff
@@ -149,69 +151,82 @@ void Run(const string& pose_graph_filename, const string& bag_filename,
...
-  CHECK_EQ(pose_graph_proto.trajectory_size(), 1)
-      << "Only pose graphs containing a single trajectory are supported.";
-  const carto::mapping::proto::Trajectory& trajectory_proto =
-      pose_graph_proto.trajectory(0);
-  CHECK_GT(trajectory_proto.node_size(), 0)
-      << "Trajectory does not contain any nodes.";
+  CHECK_EQ(pose_graph_proto.trajectory_size(), bag_filenames.size())
+      << "Pose graphs contains " << pose_graph_proto.trajectory_size()
+      << " trajectories while " << bag_filenames.size()
+      << " bags where provided. This tool requires one bag for each "
```
wohe: typo: `were provided`

cartographer_ros/cartographer_ros/assets_writer_main.cc
```diff
@@ -149,69 +151,82 @@ void Run(const string& pose_graph_filename, const string& bag_filename,
...
+      const string& bag_filename = bag_filenames[trajectory_id];
+      LOG(INFO) << "Processing " << bag_filename << "...";
+      if (trajectory_proto.node_size() == 0) {
+        continue;
       }
-      if (message.isType<sensor_msgs::MultiEchoLaserScan>()) {
-        HandleMessage(*message.instantiate<sensor_msgs::MultiEchoLaserScan>(),
-                      tracking_frame, tf_buffer,
-                      *transform_interpolation_buffer, pipeline);
+      tf2_ros::Buffer tf_buffer(::ros::DURATION_MAX);
```
wohe: Should there be a `TODO` that this is known to be very inefficient and should be fixed?

SirVer: already further up in this file.

PR #396, opened 2017-06-23
Fix PCL linking issue in Zesty

### LINKED ISSUE #355: Zesty/ROS Lunar build
Hi, I was trying to build Cartographer on Ubuntu 17.04/ROS Lunar.

I stumbled upon a linking error related to a dependency of the PCL library, the MPI library. I believe this is a bug in the `libpcl-dev` package on Zesty. I described the error, including a minimal working example, on the PCL-users mailing list:
http://www.pcl-users.org/Build-failure-on-Ubuntu-17-04-td4044552.html

I also found a really dirty fix, which is to modify the system's PCL CMake module config file to remove the offending library (described in the reply to myself on the mailing list). 

After this, I managed to build Cartographer without problems :-) Maybe you guys could chime in, I'm wondering where else should I report this (I posted a bug on [Launchpad](https://bugs.launchpad.net/ubuntu/+source/pcl/+bug/1694397)).

ojura:
Fixes #355.

### DIFF
--- a/cartographer_ros/CMakeLists.txt
+++ b/cartographer_ros/CMakeLists.txt
@@ -48,6 +48,12 @@ pkg_search_module(YAMLCPP REQUIRED yaml-cpp>=0.5.1)
 
 find_package(LuaGoogle REQUIRED)
 find_package(PCL REQUIRED COMPONENTS common io)
+# Fix for PCL linking issue in Zesty, as in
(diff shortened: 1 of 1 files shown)


ojura:
Although it looks nicer when the fix is in the section where PCL is linked, I think it's better to put the fix right next to find_package()-ing PCL, because the build system is in a broken state after that line.

ojura:
This got fixed upstream, so they removed this hack from perception_pcl (https://github.com/ros-perception/perception_pcl/commit/8b32dcd0784617cb22d1856df5858eecee0b27ed). 

I just tested, Cartographer now builds on an up-to-date Zesty. Do you want to revert this?

PR #404, opened 2017-07-04
Add submap orgins to viz marker script

jihoonl:
To visualize origin of submaps in rviz. It would be nicer if we can integrate this in `cartographer_rviz` plugin later.

submap_pose

### DIFF
--- a/cartographer_ros/scripts/submap_origins_to_marker.py
+++ b/cartographer_ros/scripts/submap_origins_to_marker.py
@@ -1,0 +1,94 @@
+#!/usr/bin/env python
+# Copyright 2017 The Cartographer Authors
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


SirVer:
Why not do this in c++ in the rViz plugin? You could add a dot in the place of the origin.

jihoonl:
Yeah. I agree that it would be better to do this in rviz plugin. It is a simple workaround before implementing it in `cartographer_rviz`.  I opened this PR with the current script based approach because I don't know when I will spend a time to implement this feature in cartographer_rviz plugin.

SirVer:
Glad that we agree that the rViz plugin is the right place for this. I will close this pull request then, since it seems this forks the responsibilities of the rViz plugin and therefore muddies responsibilities.  @jihoonl You cool with that?

I opened #404 to track the c++ implementation.

jihoonl:
Yup. I am fine with it. :wink:

PR #407, opened 2017-07-05
Publish last state of cartographer before exiting in offline node

jihoonl:
So you can check the finalized submaps, trajectories, and constraints via rviz.


ojura:
See #368. It's a better idea to keep spinning (controllable by a command line option, e.g. -keep_spinning). If the offline node quits, you can't fetch submaps anymore. I can prepare a pull request for this.

jihoonl:
Ah. yes your approach would fulfill my needs too. I didn't see #368. Thanks. Closing.

Issue #332, opened 2017-05-09
Extracting trajectory/poses


struncp:
Short version: I haven't found anything adressing this topic, so: Is there a way how to get robots trajectory from Cartographer with ROS? 

Long version:
Hello, I am comparing SLAM algorithms based on their poses throughout their trajectory in my thesis and I can't come up with a way to get a trajectory from cartographer. I use .bag data with ground truth from MIT Stata center (link below). Currently I'm obtaining trajectory via hector trajectory server and it worked fine with both Hector SLAM and gMapping but Cartographers trajectory is quite off

![cartographer_offset](https://cloud.githubusercontent.com/assets/28577045/25873862/5726af2c-3510-11e7-8d23-690e5b931f68.png)

This is from run at 0.2 speed. When I run it at realtime and robot is coming back from that part "north" to the main long hall (that part that looks like two curved horns) back that long hall to go "east" it simply skips like this

![cartochybakrouzek](https://cloud.githubusercontent.com/assets/28577045/25873870/60011416-3510-11e7-9a22-8d853d930280.png)

I believe there is some hard computing when cartographer is closing the loop from that "north" part which is not compatible with hector trajectory server. I also hope that I have right settings for cartographer because map on the outcome is correct.

source code: https://www.dropbox.com/s/6mhipp0keaj1ee3/cartographer_source.rar?dl=0

used bag: 2012-01-28-11-12-01.bag.noimages (in ground truth) at http://projects.csail.mit.edu/stata/downloads.php

Thanks for any feedback in advance.

SirVer:
> I can't come up with a way to get a trajectory from cartographer.

You run Cartographer  and once all your data has been replayed, you use `rosservice call /finish_trajectory map`. It will put files in `~/.ros/map*`, one of them is the trajectory in [a protobuf](https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping/proto/trajectory.proto). Alternatively you can use the [offline tool](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/launch/offline_backpack_2d.launch) to run through your bag as quickly as possible and receive the trajectory next to the bag.

> via hector trajectory server

To me this sounds like you are comparing apples with oranges then. 

[...]

PR #408, opened 2017-07-05
Ignore empty child_frame_id in odometry message

ojura:
In case `child_frame_id` in the odometry message is empty, assume
identity transform for `sensor_to_tracking`.

Otherwise, when `TfBridge::LookupToTracking` is passed an empty string, errors occur.
Use case: [mit stata center data set](http://projects.csail.mit.edu/stata/downloads.php)

### DIFF
--- a/cartographer_ros/cartographer_ros/sensor_bridge.cc
+++ b/cartographer_ros/cartographer_ros/sensor_bridge.cc
@@ -16,6 +16,7 @@
 
 #include "cartographer_ros/sensor_bridge.h"
 
+#include "cartographer/common/make_unique.h"
 #include "cartographer_ros/msg_conversion.h"
 #include "cartographer_ros/time_conversion.h"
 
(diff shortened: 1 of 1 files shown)


SirVer:
The [documentation](http://docs.ros.org/api/nav_msgs/html/msg/Odometry.html) does not really allow for `child_frame_id` to be empty. And why should this be identity if unset?

How about fixing their data instead of adding a workaround into Cartographer?

ojura:
I submitted it because I thought that it would be a reasonable assumption that the odometry is expressed with regards to the tracking frame if this information is missing, rather than to error out. I was aware that the chances were low that you would let this slip in, I just wanted to bring it to your attention :) 

> How about fixing their data (and contribute the fixed data back to them) instead of adding a workaround into Cartographer?

I would love to do this, but these bags are _huge_. From 50 to 70 gigabytes.

I think I know how this error slipped by. Most of the ROS nodes that I have seen which do localization take the odometry information directly from the TF tree (it would be nice if Cartographer could do this too). As a consequence of its limited use, the separate odometry topic often goes missing in various datasets, or in this case, has a defect.

Feel free to close the PR; I will be keeping this commit in my fork until I am done with these bags.

SirVer:
> I was aware that the chances were low that you would let this slip in, I just wanted to bring it to your attention :)

We are actually aware that this data is broken. I saw this in https://github.com/googlecartographer/cartographer_ros/issues/332#issuecomment-302080051. 

> I think I know how this error slipped by. Most of the ROS nodes that I have seen which do localization (e.g. AMCL, GMapping) take the odometry information directly from the TF tree (it would be nice if Cartographer could do this too).

I am not very experienced about odometry and ROS, but my understanding is that this is the old way of doing odometry and that it is deprecated and should not be done anymore. Is that not the case?

ojura:
I don't consider myself a big authority on this either. But here's some more rambling: another significant difference is that the Odometry message has twist information, which is not directly available from TF (there exists `lookupTwist`, but it's just numeric differentiation within an averaging interval). Cartographer is not using the twist information (but it could, and probably will). Until then, maybe it would make sense to also support reading odometry from TF, just for interoperability's sake.

ojura:
So, we're not merging this? Again, free to close it if not.

SirVer:
Sounds to me like a node should rewrite tf into odometry data instead. This seems the wrong approach to me. Sorry :(

PR #411, opened 2017-07-05
support map loading in offline node

jihoonl:


wohe:
Not sure whether this makes sense. The offline node consumes data as quickly as possible which means it might fall behind easily and does not produce meaningful output. I think the offline node is really only useful for producing a map as quickly as possible. Opinions? @SirVer

jihoonl:
I'm considering the case of merging a new rosbag into an existing serialized data. Assume that the new rosbags discovered a new area that hasn't been visited before. Then, you just want to add a new information into the existing data instead of replaying all rosbags again.

Or do you have a plan to provide a way of merging two serialized data into one?

ojura:
This is my use case too, I would also like the ability to load prior mapping knowledge into the offline node.

SirVer:
That makes sense to me. Let's get this in then.

Note that right now going this route will not be as good as doing multi trajectory. The reason is that the serialized state does not contain the associated scans for each submap (and no IMU and odometry data of course) - which means that loop closure can only be done from new scans to old maps, not the other way around. And of course the loaded map is frozen in the optimization problem - which could be configurable as well.

ojura:
Great! Although, isn't loop closure currently done between new scans and finished submaps?

wohe:
Then maybe add a `TODO` that `LoadMap` should be replaced by some better deserialization of full SLAM state as non-frozen trajectories once possible? And change the parameter name to reflect the intention?

jihoonl:
I added `TODO` to replace `LoadMap` and renamed `map_filename` to `pbstream_filename`

ojura:
@jihoonl you don't necessarily have to claim the TODO :) The person inside the parentheses is the assigned person, who may not be the same person as the TODO author.

PR #414, opened 2017-07-06
Keep waiting before finalization in offline node

ojura:
Addresses #407 by @jihoonl, and #368 (partially).

- The custom SIGINT handler has been removed, in favor of using the standard ROS handler and `ros::ok()`.
- I have tried to use an `AsyncSpinner`, as I originally imagined in #368. This way, we could regularly run trajectory finalization and at the same time keep spinning in a separate thread. Unfortunately, I have run into a multithreading bug which causes memory corruption (if anyone is willing to help me debug this, we can set up a Google hangout one day). 
- This means we have to spin in the main thread. In order to be able to watch the final optimization progress in RViz, without spinning asynchronously, we need to spin before calling `FinishTrajectory()`, because this would block the main thread.
- I have singled out the last two commits because they are a bit hacky (especially the last one), so I expect that you will want one of them (or both) reverted. In cbce5c2, I continue publishing the clock from the last message, frozen in time. This way, e.g. if you restart RViz or hit reset, it still receives the correct clock. The last commit advances this clock with the wall rate - this is a hack to keep some things  in ROS happy (e.g. `rosrun tf view_frames`), and I fully expect you'll want me to revert this, but I'm going for it anyway :-)

### DIFF
--- a/cartographer_ros/cartographer_ros/offline_node_main.cc
+++ b/cartographer_ros/cartographer_ros/offline_node_main.cc
@@ -65,6 +65,7 @@ constexpr char kClockTopic[] = "clock";
 constexpr char kTfStaticTopic[] = "/tf_static";
 constexpr char kTfTopic[] = "tf";
 constexpr int kLatestOnlyPublisherQueueSize = 1;
+constexpr double kClockPublishFrequency = 30.;
 
 // TODO(hrapp): This is duplicated in node_main.cc. Pull out into a config
 // unit.
@@ -177,6 +178,7 @@ void Run(const std::vector<string>& bag_filenames) {
     rosbag::View view(bag);
     const ::ros::Time begin_time = view.getBeginTime();
     const double duration_in_seconds = (view.getEndTime() - begin_time).toSec();
+    rosgraph_msgs::Clock clock;
 
     // We make sure that tf_messages are published before any data messages, so
     // that tf lookups always work and that tf_buffer has a small cache size -
@@ -238,7 +240,6 @@ void Run(const std::vector<string>& bag_filenames) {
               ->HandleOdometryMessage(
                   topic, delayed_msg.instantiate<nav_msgs::Odometry>());
         }
-        rosgraph_msgs::Clock clock;
         clock.clock = delayed_msg.getTime();
         clock_publisher.publish(clock);
 
@@ -262,7 +263,12 @@ void Run(const std::vector<string>& bag_filenames) {
     bag.close();
 
     if (FLAGS_keep_running) {
-      ::ros::spin();
+      ::ros::WallRate rate(kClockPublishFrequency);
+      while (::ros::ok()) {
+        clock_publisher.publish(clock);
+        ::ros::spinOnce();
+        rate.sleep();
+      }
     }
 
     node.map_builder_bridge()->FinishTrajectory(trajectory_id);
--- a/cartographer_ros/cartographer_ros/offline_node_main.cc
+++ b/cartographer_ros/cartographer_ros/offline_node_main.cc
@@ -54,6 +54,9 @@ DEFINE_bool(use_bag_transforms, true,
             "Whether to read, use and republish the transforms from the bag.");
 DEFINE_string(pbstream_filename, "",
               "If non-empty, filename of a pbstream to load.");
+DEFINE_bool(keep_running, false,
+            "Keep running the offline node after all messages from the bag "
+            "have been processed.");
 
 namespace cartographer_ros {
 namespace {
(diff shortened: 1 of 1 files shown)


ojura:
@SirVer I have rebased on top of upstream.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -162,7 +162,7 @@ void Run(const std::vector<string>& bag_filenames) {
   }
 
   for (const string& bag_filename : bag_filenames) {
-    if (sigint_triggered) {
+    if (!::ros::ok()) {
```
SirVer: This exactly reverts 

https://github.com/googlecartographer/cartographer_ros/commit/7551d770dc3b569d95b7cde4ba0847201f24238e

@wohe Wolfgang, do you remember why you did this back then?

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -60,10 +63,7 @@ constexpr char kClockTopic[] = "clock";
 constexpr char kTfStaticTopic[] = "/tf_static";
 constexpr char kTfTopic[] = "tf";
 constexpr int kLatestOnlyPublisherQueueSize = 1;
-
-volatile std::sig_atomic_t sigint_triggered = 0;
-
-void SigintHandler(int) { sigint_triggered = 1; }
+constexpr double kClockPublishFrequency = 30.;
```
SirVer: kClockPublishFrequencyHz

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -257,6 +257,17 @@ void Run(const std::vector<string>& bag_filenames) {
     }
 
     bag.close();
+
+    if (FLAGS_keep_running) {
```
SirVer: What does this help? No more data is fed into Cartographer, so it will never do the final optimization, so your map will never converge to the final result. And in the multi trajectory case you will just spin here forever and never start on the second trajectory.

I think you have to add hooks into the MapBuilder and maybe the sparse pose graph/constraint builder to give you an update on the state while it is optimizing.

cartographer_ros/cartographer_ros/offline_node_main.cc
wohe: I can't remember, but looking at the commit I can only hypothesize that our code does not like it if ROS is terminating while Cartographer is still running, so we handle Ctrl+C ourselves so that we always have `ros::ok()` until the very end?

wohe: @ojura, what behavior are you fixing with this revert?

ojura: Sorry, I did not realize that this is a revert. I thought that you initially added your own interrupt handler (probably missed this while doing git blame), and I thought it would be more clean to use the ROS one.

cartographer_ros/cartographer_ros/offline_node_main.cc
ojura: Ouch, now I see that this approach is flawed, especially with multi trajectory.

> What does this help? No more data is fed into Cartographer, so it will never do the final optimization, so your map will never converge to the final result.

That is true. However, it still _is_ optimizing in the background, right? And if it finishes everything, trajectory finalization is effectively done, and calling it is a no-op.

> I think you have to add hooks into the MapBuilder and maybe the sparse pose graph/constraint builder to give you an update on the state while it is optimizing.

But `FinalizeTrajectory()` blocks the main thread, which should be spinning.

I would be happiest if we could just do this, launch an `AsyncSpinner` and forget about it:
https://github.com/googlecartographer/cartographer_ros/compare/master...larics:keep-waiting-async

However, I am getting crashes. Would one of you be willing to join me on Hangouts, for example on Tuesday or Wednesday next week? I have captured the crash in a reversible debugger, and I am certain that you'll spot right away what's going on.

If that fails, I'll look into expanding this approach.

SirVer: I'll be on vacation most of next week, but I think the week after we can do a hangout.

ojura: @SirVer @wohe are you available tomorrow?

wohe: If today does not work, I can move the VC to tomorrow same time? Did you get my email?

ojura: I haven't received your mail, weird. Which address have you sent it to?

ojura: I just rechecked - it seems that my University's mail system has not forwarded the message to my GMail account. This is quite alarming, I now have to recheck if I have missed any other mails :O Today seems good, see you then!

Issue #405, opened 2017-07-05
Add submap origin visualization to the rViz plugin.


SirVer:
There should probably be a checkbox to toggle this on and off.

#404 provided a workaround in Python that can be used for the moment, but the correct implementation for this is in the rViz plugin.

PR #416, opened 2017-07-06
Display submap poses in Submaps plugin

ojura:
This is still unpolished, but I thought I might share this if anyone needs it sooner (or wants to provide feedback). Addresses #405 and #404.

Todo and ideas:
 - toggle axes visibility with a checkbox
 - adjustable length/radius
 - add a selection handler
 - show submap ID next to axes, also toggleable with a checkbox
 - option to display axes even if submap is hidden (for using with the regular occupancy grid map plugin)
 - display numeric pose coordinates under an expandable menu for each submap, similar to TF display

### DIFF
--- a/cartographer_rviz/cartographer_rviz/drawable_submap.cc
+++ b/cartographer_rviz/cartographer_rviz/drawable_submap.cc
@@ -64,15 +64,21 @@ Ogre::Quaternion ToOgre(const Eigen::Quaterniond& q) {
 
 }  // namespace
 
-DrawableSubmap::DrawableSubmap(const ::cartographer::mapping::SubmapId& id,
-                               Ogre::SceneManager* const scene_manager,
-                               ::rviz::Property* submap_category,
-                               const bool visible)
+DrawableSubmap::DrawableSubmap(
+    const ::cartographer::mapping::SubmapId& id,
+    ::rviz::DisplayContext* const display_context,
+    ::rviz::Property* const submap_category, const bool visible,
+    const float pose_axes_length, const float pose_axes_radius)
     : id_(id),
-      scene_manager_(scene_manager),
-      scene_node_(scene_manager->getRootSceneNode()->createChildSceneNode()),
-      manual_object_(scene_manager->createManualObject(
+      display_context_(display_context),
+      scene_node_(display_context->getSceneManager()
+                      ->getRootSceneNode()
+                      ->createChildSceneNode()),
+      submap_node_(scene_node_->createChildSceneNode()),
+      manual_object_(display_context->getSceneManager()->createManualObject(
           kManualObjectPrefix + GetSubmapIdentifier(id))),
+      pose_axes_(display_context->getSceneManager(), scene_node_,
+                 pose_axes_length, pose_axes_radius),
       last_query_timestamp_(0) {
   material_ = Ogre::MaterialManager::getSingleton().getByName(
       kSubmapSourceMaterialName);
@@ -90,9 +96,8 @@ DrawableSubmap::DrawableSubmap(const ::cartographer::mapping::SubmapId& id,
   visibility_ = ::cartographer::common::make_unique<::rviz::BoolProperty>(
       "" /* title */, visible, "" /* description */, submap_category,
       SLOT(ToggleVisibility()), this);
-  if (visible) {
-    scene_node_->attachObject(manual_object_);
-  }
+  submap_node_->attachObject(manual_object_);
+  scene_node_->setVisible(visible);
   connect(this, SIGNAL(RequestSucceeded()), this, SLOT(UpdateSceneNode()));
 }
 
@@ -107,8 +112,9 @@ DrawableSubmap::~DrawableSubmap() {
     Ogre::TextureManager::getSingleton().remove(texture_->getHandle());
     texture_.setNull();
   }
-  scene_manager_->destroySceneNode(scene_node_);
-  scene_manager_->destroyManualObject(manual_object_);
+  display_context_->getSceneManager()->destroySceneNode(submap_node_);
+  display_context_->getSceneManager()->destroySceneNode(scene_node_);
+  display_context_->getSceneManager()->destroyManualObject(manual_object_);
 }
 
 void DrawableSubmap::Update(
@@ -250,8 +256,11 @@ void DrawableSubmap::UpdateTransform() {
   CHECK(submap_texture_ != nullptr);
   const ::cartographer::transform::Rigid3d pose =
       pose_ * submap_texture_->slice_pose;
-  scene_node_->setPosition(ToOgre(pose.translation()));
-  scene_node_->setOrientation(ToOgre(pose.rotation()));
+  submap_node_->setPosition(ToOgre(pose.translation()));
+  submap_node_->setOrientation(ToOgre(pose.rotation()));
+  pose_axes_.setPosition(ToOgre(pose_.translation()));
+  pose_axes_.setOrientation(ToOgre(pose_.rotation()));
+  display_context_->queueRender();
 }
 
 float DrawableSubmap::UpdateAlpha(const float target_alpha) {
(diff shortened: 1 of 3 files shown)


cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -138,8 +138,8 @@ void SubmapsDisplay::processMessage(
       trajectory.emplace(
           submap_index,
           ::cartographer::common::make_unique<DrawableSubmap>(
-              trajectory_id, submap_index, context_->getSceneManager(),
-              trajectory_category.get(), visibility_all_enabled_->getBool()));
+              trajectory_id, submap_index, context_, trajectory_category.get(),
+              visibility_all_enabled_->getBool(), 0.3f, 0.06f));
```
SirVer: will we ever want to change these? If so, they should be configurable, if not they should be constexpr next to where they are used. I put my vote in for the second option for now.

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -70,6 +72,10 @@ class DrawableSubmap : public QObject {
   // 'current_tracking_z'.
   void SetAlpha(double current_tracking_z);
 
+  // todo
```
SirVer: they are not implemented. Remove for now and add back when they are actually added?

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -95,11 +101,14 @@ class DrawableSubmap : public QObject {
...
 
   ::cartographer::common::Mutex mutex_;
+  ::rviz::DisplayContext* const context_;
   Ogre::SceneManager* const scene_manager_;
   Ogre::SceneNode* const scene_node_;
+  Ogre::SceneNode* const submap_node_;
   Ogre::ManualObject* manual_object_;
   Ogre::TexturePtr texture_;
   Ogre::MaterialPtr material_;
+  ::rviz::Axes axes_;
```
SirVer: how about `origin_marker_`?

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -46,9 +47,10 @@ class DrawableSubmap : public QObject {
  public:
   // Each submap is identified by a 'trajectory_id' plus a 'submap_index'.
   // 'scene_manager' is the Ogre scene manager to which to add a node.
-  DrawableSubmap(int trajectory_id, int submap_index,
-                 Ogre::SceneManager* scene_manager,
-                 ::rviz::Property* submap_category, const bool visible);
+  DrawableSubmap(const int trajectory_id, const int submap_index,
```
SirVer: const int -> int. The const is an implementation detail for copied and has no meaning in the interface, which (ideally) should be the only thing exposed in the header.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -266,15 +273,7 @@ float DrawableSubmap::UpdateAlpha(const float target_alpha) {
...
-  if (visibility_->getBool()) {
-    if (scene_node_->numAttachedObjects() == 0) {
-      scene_node_->attachObject(manual_object_);
-    }
-  } else {
-    if (scene_node_->numAttachedObjects() > 0) {
-      scene_node_->detachObject(manual_object_);
-    }
-  }
+  scene_node_->setVisible(visibility_->getBool());
```
SirVer: Why this change? I vaguely remember you had the longer code before for some performance reasons or so?

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -95,11 +101,14 @@ class DrawableSubmap : public QObject {
   const int submap_index_;
 
   ::cartographer::common::Mutex mutex_;
+  ::rviz::DisplayContext* const context_;
   Ogre::SceneManager* const scene_manager_;
   Ogre::SceneNode* const scene_node_;
+  Ogre::SceneNode* const submap_node_;
```
SirVer: I see that the submap_node_ is made a child node of the scene_node_, but why is the scene_node_ not enough?

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -55,16 +55,20 @@ std::string GetSubmapIdentifier(const int trajectory_id,
...
+                               ::rviz::Property* const submap_category,
+                               const bool visible, const float axes_length,
+                               const float axes_radius)
     : trajectory_id_(trajectory_id),
       submap_index_(submap_index),
-      scene_manager_(scene_manager),
-      scene_node_(scene_manager->getRootSceneNode()->createChildSceneNode()),
-      manual_object_(scene_manager->createManualObject(
+      context_(context),
+      scene_manager_(context->getSceneManager()),
```
SirVer: delete member? Now you have the context_, you can always get the sceneManager from it?

SirVer:
You are on fire again I see :). Thanks!

Comments are in code.

jihoonl:
Awesome thanks @ojura :smile: :smile: :smile:

cartographer_rviz/cartographer_rviz/drawable_submap.h
ojura: It's not a ROS marker. Maybe `pose_axes_`?

cartographer_rviz/cartographer_rviz/drawable_submap.cc
ojura: Nah, I just didn't know about `setVisible()`. Plus, now I can handle visibility of both axes and the submap by having them both under a signle `scene_node_`.

cartographer_rviz/cartographer_rviz/drawable_submap.h
ojura: See `UpdateTransform()`. `submap_node_` (ex `scene_node_`) has its pose set to the slice pose; I would have to perform additional transformations to obtain the submap origin in the coordinate system of the submap slice. By having a `scene_node_` which has the same pose as the root node, I can just set the axes pose to the one written in `position_` and `orientation_`. Plus, I can hide both the submap and the axes by calling `scene_node_.setVisible()`.

cartographer_rviz/cartographer_rviz/submaps_display.cc
ojura: Yes, I will add an RViz property for it.

cartographer_rviz/cartographer_rviz/drawable_submap.h
ojura: Okay. And the boolean variable retains the const in the header?

SirVer: No, that was wrong already. Could you fix it?

cartographer_rviz/cartographer_rviz/submaps_display.cc
SirVer: next PR?

cartographer_rviz/cartographer_rviz/drawable_submap.h
SirVer: There is no such thing as a too small PR. Next PRs?

cartographer_rviz/cartographer_rviz/drawable_submap.h
SirVer: fair enough. pose_axes_ is a good name.

cartographer_rviz/cartographer_rviz/drawable_submap.h
ojura: Okay, sure. The header file is full of consts though. Should I remove them all?

SirVer: I did that real quick in this branch while reviewing.

SirVer:
@ojura I rebased while having another quick look over this.

cartographer_rviz/cartographer_rviz/drawable_submap.h
ojura: Sure. The rviz properties pointers should also have the `const` removed, right?

SirVer: Yes... I only did half of the job :(

ojura:
@SirVer okay, I addressed the nits, it's not yet finished (I added a TODO for making it configurable from RViz) but it should be ready for merging :)

ojura:
Rebased to follow #419.

ojura:
Sure.

SirVer:
Thanks for your work and sorry that this took so long on my side.

PR #422, opened 2017-07-12
Toggleable trajectory

jihoonl:
To enable or disable visibility per trajectory. I introduced drawable_trajectory object for per trajectory object management.

toggleable_trajectory

### DIFF
--- a/cartographer_rviz/cartographer_rviz/submaps_display.cc
+++ b/cartographer_rviz/cartographer_rviz/submaps_display.cc
@@ -107,9 +107,10 @@ void SubmapsDisplay::processMessage(
     if (trajectory_id >= trajectories_.size()) {
       continue;
     }
-    const auto& trajectory = trajectories_[trajectory_id].second;
-    const auto it = trajectory.find(submap_entry.submap_index);
-    if (it != trajectory.end() &&
+
+    const auto& submaps = trajectories_[trajectory_id]->submaps();
+    const auto it = submaps.find(submap_entry.submap_index);
+    if (it != submaps.end() &&
         it->second->version() > submap_entry.submap_version) {
       // Versions should only increase unless Cartographer restarted.
       trajectories_.clear();
@@ -122,37 +123,30 @@ void SubmapsDisplay::processMessage(
     const SubmapId id{submap_entry.trajectory_id, submap_entry.submap_index};
     listed_submaps.insert(id);
     while (id.trajectory_id >= static_cast<int>(trajectories_.size())) {
-      trajectories_.push_back(Trajectory(
-          ::cartographer::common::make_unique<::rviz::Property>(
-              QString("Trajectory %1").arg(id.trajectory_id), QVariant(),
-              QString("List of all submaps in Trajectory %1.")
-                  .arg(id.trajectory_id),
-              submaps_category_),
-          std::map<int, std::unique_ptr<DrawableSubmap>>()));
+      trajectories_.push_back(
+          ::cartographer::common::make_unique<PerTrajectorySubmapDisplay>(
+              id.trajectory_id, submaps_category_, context_,
+              visibility_all_enabled_->getBool()));
     }
-    auto& trajectory_category = trajectories_[id.trajectory_id].first;
-    auto& trajectory = trajectories_[id.trajectory_id].second;
-    if (trajectory.count(id.submap_index) == 0) {
-      // TODO(ojura): Add RViz properties for adjusting submap pose axes
-      constexpr float kSubmapPoseAxesLength = 0.3f;
-      constexpr float kSubmapPoseAxesRadius = 0.06f;
-      trajectory.emplace(id.submap_index,
-                         ::cartographer::common::make_unique<DrawableSubmap>(
-                             id, context_, trajectory_category.get(),
-                             visibility_all_enabled_->getBool(),
-                             kSubmapPoseAxesLength, kSubmapPoseAxesRadius));
+
+    auto& trajectory = trajectories_[id.trajectory_id];
+    const auto& submaps = trajectory->submaps();
+    if (submaps.count(id.submap_index) == 0) {
+      trajectory->AddSubmap(id);
     }
-    trajectory.at(id.submap_index)
-        ->Update(msg->header, submap_entry, context_->getFrameManager());
+    trajectory->UpdateSubmap(id.submap_index, msg->header, submap_entry,
+                             context_->getFrameManager());
   }
+
   // Remove all submaps not mentioned in the SubmapList.
   for (size_t trajectory_id = 0; trajectory_id < trajectories_.size();
        ++trajectory_id) {
-    auto& trajectory = trajectories_[trajectory_id].second;
-    for (auto it = trajectory.begin(); it != trajectory.end();) {
+    auto& trajectory = trajectories_[trajectory_id];
+    auto& submaps = trajectory->submaps();
+    for (auto it = submaps.begin(); it != submaps.end();) {
       if (listed_submaps.count(
               SubmapId{static_cast<int>(trajectory_id), it->first}) == 0) {
-        it = trajectory.erase(it);
+        it = submaps.erase(it);
       } else {
         ++it;
       }
@@ -169,7 +163,8 @@ void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
                                    tracking_frame_property_->getStdString(),
                                    ros::Time(0) /* latest */);
     for (auto& trajectory : trajectories_) {
-      for (auto& submap_entry : trajectory.second) {
+      auto& submaps = trajectory->submaps();
+      for (auto& submap_entry : submaps) {
         submap_entry.second->SetAlpha(
             transform_stamped.transform.translation.z);
       }
@@ -181,13 +176,14 @@ void SubmapsDisplay::update(const float wall_dt, const float ros_dt) {
   // Schedule fetching of new submap textures.
   for (const auto& trajectory : trajectories_) {
     int num_ongoing_requests = 0;
-    for (const auto& submap_entry : trajectory.second) {
+    const auto& submaps = trajectory->submaps();
+    for (const auto& submap_entry : submaps) {
       if (submap_entry.second->QueryInProgress()) {
         ++num_ongoing_requests;
       }
     }
-    for (auto it = trajectory.second.rbegin();
-         it != trajectory.second.rend() &&
+    for (auto it = submaps.rbegin();
+         it != submaps.rend() &&
          num_ongoing_requests < kMaxOnGoingRequestsPerTrajectory;
          ++it) {
       if (it->second->MaybeFetchTexture(&client_)) {
@@ -201,9 +197,7 @@ void SubmapsDisplay::AllEnabledToggled() {
   ::cartographer::common::MutexLocker locker(&mutex_);
   const bool visibility = visibility_all_enabled_->getBool();
   for (auto& trajectory : trajectories_) {
-    for (auto& submap_entry : trajectory.second) {
-      submap_entry.second->set_visibility(visibility);
-    }
+    trajectory->set_visibility(visibility);
   }
 }
 
--- a/cartographer_rviz/cartographer_rviz/submaps_display.cc
+++ b/cartographer_rviz/cartographer_rviz/submaps_display.cc
@@ -108,15 +108,12 @@ void SubmapsDisplay::processMessage(
       continue;
     }
 
-    const auto& submaps = trajectories_[trajectory_id]->submaps();
-    const auto it = submaps.find(submap_entry.submap_index);
-    if (it != submaps.end() &&
-        it->second->version() > submap_entry.submap_version) {
-      // Versions should only increase unless Cartographer restarted.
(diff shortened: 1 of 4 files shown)


cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
```diff
@@ -0,0 +1,63 @@
...
+
+DrawableTrajectory::DrawableTrajectory(const int trajectory_id,
+                                       ::rviz::Property* submap_category,
+                                       Ogre::SceneManager* scene_manager,
+                                       const bool visible)
+    : id_(trajectory_id),
+      submaps_category_(submap_category),
+      scene_manager_(scene_manager),
+      submaps_() {
+  property_ = new ::rviz::BoolProperty(
```
ojura: Maybe call this field `visibility`, to follow `drawable_submap.cc`? Also, use a unique pointer (with `::cartographer::common::make_unique`) instead of a raw pointer with new/delete()?

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
```diff
@@ -0,0 +1,63 @@
...
+      QString("Trajectory %1").arg(id_), visible,
+      QString("List of all submaps in Trajectory %1.").arg(id_),
+      submaps_category_, SLOT(AllEnabledToggled()), this);
+}
+
+DrawableTrajectory::~DrawableTrajectory() {
+  submaps_.clear();
```
ojura: Doesn't the vector's destructor already do this?

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -107,49 +107,48 @@ void SubmapsDisplay::processMessage(
...
-    auto& trajectory = trajectories_[id.trajectory_id].second;
-    if (trajectory.count(id.submap_index) == 0) {
-      trajectory.emplace(
-          id.submap_index,
-          ::cartographer::common::make_unique<DrawableSubmap>(
-              id, context_->getSceneManager(), trajectory_category.get(),
+      trajectories_.push_back(
+          ::cartographer::common::make_unique<DrawableTrajectory>(
+              id.trajectory_id, submaps_category_, context_->getSceneManager(),
               visibility_all_enabled_->getBool()));
```
ojura: I like the hierarchy: the global visibility enabled toggle controls the visibility of new trajectories, while the per-trajectory toggle controls visibility of newly inserted submaps for each trajectory. Good job :)

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
```diff
@@ -0,0 +1,63 @@
+/*
+ * Copyright 2016 The Cartographer Authors
```
ojura: Maybe 2017 here? :D

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
```diff
@@ -0,0 +1,63 @@
...
+ */
+
+#include "cartographer_rviz/drawable_trajectory.h"
+
+#include "cartographer/common/make_unique.h"
+
+namespace cartographer_rviz {
+
+DrawableTrajectory::DrawableTrajectory(const int trajectory_id,
+                                       ::rviz::Property* submap_category,
```
SirVer: consistent naming: `submap_category` vs `submaps_category_`.

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
SirVer: Meh... I am really unsure how to correctly handle this. Maybe we can just remove the years in the copyright headers eventually. For now, just keep it 2016.

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
```diff
@@ -0,0 +1,63 @@
...
+                                       Ogre::SceneManager* scene_manager,
+                                       const bool visible)
+    : id_(trajectory_id),
+      submaps_category_(submap_category),
```
SirVer: submaps_category_ is unused. remove?

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
```diff
@@ -0,0 +1,63 @@
...
+      scene_manager_(scene_manager),
+      submaps_() {
```
SirVer: no need to initialize a std::map, it will be default-empty.

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
SirVer: Make property_ a non-pointer member and initialize it in the initializer list No need to delete it in the destructor then. Also rename it to visible_?

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
SirVer: Yup, it does. Destructor can be empty.

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
```diff
@@ -0,0 +1,63 @@
...
+      ::cartographer::common::make_unique<DrawableSubmap>(
+          submap_id, scene_manager_, property_, property_->getBool()));
+}
+
+void DrawableTrajectory::UpdateSubmap(
+    const int submap_index, const ::std_msgs::Header& header,
+    const ::cartographer_ros_msgs::SubmapEntry& metadata,
+    ::rviz::FrameManager* frame_manager) {
+  submaps_.at(submap_index)->Update(header, metadata, frame_manager);
+}
```
SirVer: empty line and `// namespace cartographer_ros` comment. Clang-format should add that for you. Did you run it?

cartographer_rviz/cartographer_rviz/drawable_trajectory.h
```diff
@@ -0,0 +1,70 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CARTOGRAPHER_RVIZ_SRC_DRAWABLE_TRAJECTORY_H_
+#define CARTOGRAPHER_RVIZ_SRC_DRAWABLE_TRAJECTORY_H_
+
+#include <future>
+
+#include "rviz/display_context.h"
```
SirVer: only one header block, sorted alphabetically.

cartographer_rviz/cartographer_rviz/drawable_trajectory.h
```diff
@@ -0,0 +1,70 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CARTOGRAPHER_RVIZ_SRC_DRAWABLE_TRAJECTORY_H_
+#define CARTOGRAPHER_RVIZ_SRC_DRAWABLE_TRAJECTORY_H_
+
+#include <future>
```
SirVer: unused? remove.

cartographer_rviz/cartographer_rviz/drawable_trajectory.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include "rviz/display_context.h"
+#include "rviz/frame_manager.h"
+#include "rviz/properties/bool_property.h"
+
+#include "cartographer/mapping/id.h"
+#include "cartographer_rviz/drawable_submap.h"
+
+namespace cartographer_rviz {
```
SirVer: empty line after namespace.

cartographer_rviz/cartographer_rviz/drawable_trajectory.h
```diff
@@ -0,0 +1,70 @@
...
+class DrawableTrajectory : public QObject {
+  Q_OBJECT
+
+ public:
+  DrawableTrajectory(const int trajectory_id, ::rviz::Property* submap_category,
```
SirVer: const for copied values (i.e. `const int foo, const bool foo`) is meaningless information in the interface. I know we have it else where as well, but that is not correct. Please remove here.

cartographer_rviz/cartographer_rviz/drawable_trajectory.h
```diff
@@ -0,0 +1,70 @@
...
+
+#include "rviz/display_context.h"
+#include "rviz/frame_manager.h"
+#include "rviz/properties/bool_property.h"
+
+#include "cartographer/mapping/id.h"
+#include "cartographer_rviz/drawable_submap.h"
+
+namespace cartographer_rviz {
+class DrawableTrajectory : public QObject {
```
SirVer: Misleading name, since it contains Submaps and does not really draw the trajectory. PerTrajectorySubmapDisplay?

cartographer_rviz/cartographer_rviz/drawable_trajectory.h
```diff
@@ -0,0 +1,70 @@
...
+  DrawableTrajectory(const int trajectory_id, ::rviz::Property* submap_category,
+                     Ogre::SceneManager* scene_manager, const bool visible);
+  ~DrawableTrajectory() override;
+  DrawableTrajectory(const DrawableTrajectory&) = delete;
+  DrawableTrajectory& operator=(const DrawableTrajectory&) = delete;
+
+  bool visibility() const { return property_->getBool(); }
+  void set_visibility(const bool visibility) { property_->setBool(visibility); }
+
+  const std::map<int, std::unique_ptr<DrawableSubmap>>& submaps() const {
```
SirVer: This is inconsistent - you are handing out the submaps() to check various conditions, i.e. how many requests are in-flight, did Cartographer reset, did the trajectory vanish. But you also hand out a mutable version of your vector that SubmapsList deletes stuff out of.

On the other site you have `UpdateSubmap` that just calls through to the submap.

I see two possible design directions here:

1) Either make this class really own the submaps - never hand them out. Add `NumOngoingRequests`, `SetAlpha`, `processMessage` which should get rid of `AddSubmap` and `UpdateSubmap`. Change SubmapsDisplay to delegate the work to this class.
2) Alternatively (and probably easier and shorter): change SubmapsDisplay to contain a private `struct Trajectory` that is no longer just a `std::pair` and holds the boolean property you want here as well. Then handle all the logic in SubmapsDisplay.

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -107,49 +107,48 @@ void SubmapsDisplay::processMessage(
...
        ++trajectory_id) {
-    auto& trajectory = trajectories_[trajectory_id].second;
-    for (auto it = trajectory.begin(); it != trajectory.end();) {
+    auto& trajectory = trajectories_[trajectory_id];
+    auto& submaps = trajectory->submaps();
+    for (auto it = submaps.begin(); it != submaps.end();) {
       if (listed_submaps.count(
               SubmapId{static_cast<int>(trajectory_id), it->first}) == 0) {
-        it = trajectory.erase(it);
+        it = submaps.erase(it);
```
SirVer: This is iffy... You are modifying data that you are not owning. This violates separation of concerns and is really hard to reason about.

jihoonl:
Yup, I knew that the current design contains some flaws since it was touching rviz plugin first time. I wanted to verify if you agree with the structure and direction of development before making it complete. I will address the points from @SirVer and @ojura after #416 get merged.

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
jihoonl: Destructor doesn't take care of cleaning. It crashes without explicit calling.

cartographer_rviz/cartographer_rviz/drawable_trajectory.h
jihoonl: I actually tried the alternative option before introducing `PerTrajectory` Class. It was somewhat difficult to manage QT event trigger for individual trajectory with my current QT knowledge. Thus, I decided to do option 1.

cartographer_rviz/cartographer_rviz/drawable_trajectory.cc
ojura: Sounds to me something else is wrong. A map (sorry, not a vector as I stated above) will definitely destroy all the objects contained inside when it's being destroyed.

jihoonl:
Ok. This is ready again.

ojura:
Hey guys, allow me to crash the party :-) Here's an alternate, somewhat simpler implementation, which follows @SirVer's second suggestion from [this comment](https://github.com/googlecartographer/cartographer_ros/pull/422#discussion_r128189575):

https://github.com/googlecartographer/cartographer_ros/compare/master...larics:toggleable-trajectory

jihoonl:
I don't really care as long as the feature lands in upstream.

SirVer:
@jihoonl I opened and merged #456 because it is shorter and seems easier to understand to me, giving you credit. Thanks for the feature idea and implementation!

PR #434, opened 2017-07-20
Simplify sensor subscribers.

wohe:


cartographer_ros/cartographer_ros/node.h
```diff
@@ -72,6 +74,9 @@ class Node {
   bool HandleWriteAssets(
       cartographer_ros_msgs::WriteAssets::Request& request,
       cartographer_ros_msgs::WriteAssets::Response& response);
+  std::unordered_set<string> ComputeExpectedTopics(
```
SirVer: Comment?

cartographer_ros/cartographer_ros/node.h
```diff
@@ -104,12 +109,8 @@ class Node {
...
 
   // These are keyed with 'trajectory_id'.
-  std::unordered_map<int, ::ros::Subscriber> laser_scan_subscribers_;
-  std::unordered_map<int, ::ros::Subscriber> multi_echo_laser_scan_subscribers_;
-  std::unordered_map<int, ::ros::Subscriber> odom_subscribers_;
-  std::unordered_map<int, ::ros::Subscriber> imu_subscribers_;
-  std::unordered_map<int, std::vector<::ros::Subscriber>>
-      point_cloud_subscribers_;
+  std::unordered_map<int, std::vector<::ros::Subscriber>> subscribers_;
+  std::unordered_set<std::string> topics_in_use_;
```
SirVer: `subscribed_topics_`?

ojura:
A quick question. Why do you need multiple topics for scans? Isn't the frame ID of the sensor enough to distinguish them?

wohe:
What benefit do you see in not having multiple topics? I can only see disadvantages not being able to easily remap and distinguish different sensors.

ojura:
Having multiple _source_ topics of course makes sense - when you have multiple rangefinder driver nodes, each publishing scans from its rangefinder. But I don't see any advantage to having more than one _sink_ topic in the Cartographer node itself. You can always route messages from multiple topics into one topic (e.g. you would remap each of individual laser driver nodes' output topics to Cartographer's scan topic). As I said, within the Cartographer node, you can always distinguish the sensors based on the sensor frame. I just don't see what you can accomplish by having multiple sink (input to Cartographer) topics.

I am not a ROS authority by any means, but I have seen this design in e.g. GMapping and it kind of makes sense. I think it's common in ROS that sensors are differentiated by their frame IDs and not by topic names.

SirVer:
I agree, this seems to make sense to me too. It is not a completely straight forward refactoring since we rely on the topics for identification of individual devices - we'd need to switch this to frame_ids. The benefit would be that we do not need to tell the number of devices we expect which would be kinda nice.

ojura:
Ok, it's good we're on the same page :) For now maybe we can just open an issue about this, since it's not high priority. Do you want me to do it?

SirVer:
@ojura I retract my statement from above :)

I just discussed [this case here](
https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/launch/backpack_3d.launch#L24) with @wohe and now I am convinced that what we have is the best we can do. 

> Ok, it's good we're on the same page :)

Yes, but I think we were both on the wrong page, i.e. having both the same wrong mental model. topic remapping is not a pipeline, it is actually just a search & replace in incoming topics. So, if we hardecode `laser` as topic in Cartographer, you can remap `laser` -> `horizontal_laser` in the launch file, but you cannot remap both `laser` -> `horizontal_laser` and `laser` -> `vertical_laser`, i.e. it is not a pipeline.

ojura:
That's true. But I was referring to output remapping - whoever publishes `horizontal_laser` can be remapped to `laser`, and whoever publishes `vertical_laser` to also `laser`.

wohe:
Can you explain which advantage you see in doing this? As I said I can see disadvantages:

- more complicated remapping: you can no longer remap on the Cartographer side but must do it on the sensor side which can be cumbersome, e.g. when used with includes in the launch file
- the merge sorting must know which data to expect, so we have to somehow explain separately which frame IDs are expected
- we have data where different sensors have already be transformed to the same frame, to handle this we would have to artificially remap the frames to make sure different sensors can be identified

These are disadvantages. If we change the code in a way that we have to accept those, what benefits do we get? Otherwise this seems to me to make the code strictly worse.

ojura:
I don't see a huge issue in 1). I think that making heavy use of topic remapping is standard in ROS and is preferred over adapting node code (e.g. we always prefer using hardcoded and well-defined topic names, along with topic remapping and ROS namespaces, over using configurable/dynamic topic names in node code).

For 2): why couldn't the used frame IDs be updated dynamically? As long we can look up the transform from some new, previously unseen sensor frame to the tracking frame, we should be open to receiving and making use of that sensor data. As @SirVer mentioned, this would free us from having to pre-configure the number of devices.

For 3): again, I emphasize that I am not a ROS authority and everything I argue here is based on a feeling of what I have seen so far in ROS. That said, I think that a dataset like this would not be conforming to standards. [We already had a](https://github.com/googlecartographer/cartographer_ros/pull/408) case where a dataset invalidly had an empty `frame_id` for odometry sensor messages and the conclusion was that we should not bend over and do unstandard things in ROS so that these invalid datasets work out of box.

Perhaps we could ping someone from the ROS community about this?

jihoonl:
I don't see why we have to have just one subscriber for each sensor type. Having one topic for one sensor data is very natural in ROS. I don't see any benefit by integrating all source into one. 

1) It makes harder to introspect each sensor data. It is harder to use other ros introspection tools like `rviz` and `rostopic` too.

2) Sensors are not just for cartographer. Consider that laser scan and pointcloud are used for navigation, perception and other purposes. What if other nodes only want to use `horizontal_laser`?

3) Your suggestion is still possible with the current implementation already. Please don't limit its usability. I think it is oversimplification just for cartographer.

ojura:
Yeah, you're right.

> Having one topic for one sensor data is very natural in ROS.

Still, I am wondering if there's any official specs on this.

PR #443, opened 2017-07-25
Make Node thread safe

ojura:
Accessing the MapBuilderBridge in Node was not previously guarded by a mutex, hence not thread safe.

We concluded yesterday that the Node class should be refactored to include pass-through methods which would lock the Node mutex before invoking the appropriate methods of the map builder bridge. I have come up with a more elegant way to protect access to the map builder bridge, without modifying the rest of the Cartographer codebase, and without duplicating the map builder bridge interface in the Node class, which I think is pretty cool. It's based on the [execution-around-pointer C++ idiom](https://en.wikibooks.org/wiki/More_C%2B%2B_Idioms/Execute-Around_Pointer). Please have a look at https://github.com/googlecartographer/cartographer/compare/master...larics:mutex-proxy.

Here are some measures of the offline node to see how much this has impacted performance (average of 10-30 measurements):

RViz open, submaps and trajectory visualization active: 172.5 s before, 175.01 s after
RViz closed: 137.7 s before, 140.6 s after

Do you think this is an acceptable performance loss?

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -106,7 +106,10 @@ Node::~Node() {}
 
 ::ros::NodeHandle* Node::node_handle() { return &node_handle_; }
 
-MapBuilderBridge* Node::map_builder_bridge() { return &map_builder_bridge_; }
+::cartographer::common::Mutex::Proxy<MapBuilderBridge>
+Node::map_builder_bridge() {
+  return {&map_builder_bridge_, &mutex_};
+}
 
 bool Node::HandleSubmapQuery(
     ::cartographer_ros_msgs::SubmapQuery::Request& request,
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -455,7 +455,9 @@ void Node::FinishAllTrajectories() {
 
 void Node::FinishTrajectory(const int trajectory_id) {
   carto::common::MutexLocker lock(&mutex_);
+  CHECK(is_active_trajectory_.at(trajectory_id));
   map_builder_bridge_.FinishTrajectory(trajectory_id);
+  is_active_trajectory_[trajectory_id] = false;
 }
 
 void Node::HandleOdometryMessage(const int trajectory_id,
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -106,11 +106,6 @@ Node::~Node() {}
 
 ::ros::NodeHandle* Node::node_handle() { return &node_handle_; }
 
-::cartographer::common::Mutex::Proxy<MapBuilderBridge>
-Node::map_builder_bridge() {
-  return {&map_builder_bridge_, &mutex_};
-}
-
 bool Node::HandleSubmapQuery(
     ::cartographer_ros_msgs::SubmapQuery::Request& request,
     ::cartographer_ros_msgs::SubmapQuery::Response& response) {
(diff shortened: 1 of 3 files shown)


SirVer:
The fundamental problem with our race-condition bug is that `MapBuilderBridge` should not be handed out. It is owned by node and the node must protect it against multi-threaded access. 

We should move towards modeling this ownership correctly - your suggestion moves us into the different direction. Instead of just handing out data, we now also hand out the lock to this data to our users. They can hold it for as long as they like, potentially blocking the program. You are assuming that they only call a function on your proxy and delete the lock again, but your API cannot ensure this, therefore it is inviting bugs.  The problems here are very similar to the issues associated with locking on `this` in Java.

While the approach we discussed yesterday will be more code to type, it is trivial code (pure composition). It will easier express the ownership and locking semantics of the involved classes. It also moves us towards the right direction of ownership.

ojura:
> You are assuming that they only call a function on your proxy and delete the lock again, but your API cannot ensure this, therefore it is inviting bugs.

How about deleting the copy/move constructor/assignment operator of the `Proxy`? This ensures that the user cannot copy or move the returned Proxy instance, and keep the mutex for themselves - it can only shortly exist as an rvalue which allows calling methods on it.

SirVer:
Leaking the mutex is leaking an implementation detail. This will not allow us to ever move towards finer grained locking or other refactorings without changing all users.

ojura:
Sure, if you think that duplicating the interface in `Node` is the better route, we can do that. 

Although I would disagree in two things. First, that we are leaking implementation - the mutex lock (not the mutex itself) is kept as a private variable in `Proxy`, which is more or less equivalent to constructing it inside a pass-through method in `Node` - it is inaccessible to the caller.

I agree that we should eventually move towards more fine-grained locking, i.e. making `MapBuilderBridge` itself thread-safe, so that `Node` doesn't have to concern itself over locking its mutex when its `MapBuilderBridge` is being accessed. I envisioned when this is accomplished, we can just swap the return type of `map_builder_bridge()` from `Proxy` back to a regular pointer again - and you don't have to modify anything in the call sites, so I disagree that we would have to change all users (we don't have to change anything!)

The main benefit of having pass-through methods in `Node` is that it would be easier to do additional Node-related stuff before and after invoking the member methods - if you're concerned about that, sure, we can duplicate the interface, but I would defer that until it we really need it.

ojura:
Okay, I might have to eat my own words now. I have just realized that, although I disabled moving and copying of the `Proxy` object, the user can bind an rvalue reference or a const lvalue reference to the temporary object, which extends its lifetime to whole scope of the caller, and delays the destruction i.e. 
release of the mutex. So yes, you could steal the mutex lock if you try hard enough. (Maybe we could call it a feature? If you do this intentionally, you can perform method calls atomically this way :-) )

ojura:
Node refactoring is in this branch: https://github.com/googlecartographer/cartographer_ros/compare/master...larics:node-passthrough-thread-safe

I can update this pull request to contain these changes, just let me know.

SirVer:
Please update this branch.

ojura:
Done.

ojura:
@wohe Done, please have a look.

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -459,6 +457,55 @@ void Node::FinishAllTrajectories() {
   }
 }
 
+void Node::FinishTrajectory(const int trajectory_id) {
+  carto::common::MutexLocker lock(&mutex_);
+  map_builder_bridge_.FinishTrajectory(trajectory_id);
```
wohe: `CHECK` that `is_active_trajectory_` is `true` before calling `FinishTrajectory()` and set to `false` afterwards?

ojura: Ok, done.

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -459,7 +459,10 @@ void Node::FinishAllTrajectories() {
 
 void Node::FinishTrajectory(const int trajectory_id) {
   carto::common::MutexLocker lock(&mutex_);
+  CHECK(is_active_trajectory_.count(trajectory_id) > 0 &&
```
wohe: The first check is redundant. You'd also need to rebase once more before we can merge.

ojura: If it doesn't exist, won't operator[] implicitly insert a new default value (which happens to be `false` for booleans)? Do you want that?

wohe: Yes, that's what I meant. The `CHECK` will fail in this case, so it is irrelevant whether we added a new value to the map.

wohe: And if you want to better document that we expect the value to already exist you could write it like this:
```
CHECK(is_active_trajectory_.at(trajectory_id));
```

ojura: Yes, that's exactly what I had in mind, I just forgot about it. Rebased and changed to use at().

PR #452, opened 2017-07-27
use trajectory id as namespace for trajectory marker

jihoonl:
So, each trajectory can be on or off.

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -175,6 +175,7 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodeList() {
        ++trajectory_id) {
     const auto& single_trajectory_nodes = all_trajectory_nodes[trajectory_id];
     visualization_msgs::Marker marker;
+    marker.ns = std::to_string(trajectory_id);
     marker.id = marker_id++;
     marker.type = visualization_msgs::Marker::LINE_STRIP;
     marker.header.stamp = ::ros::Time::now();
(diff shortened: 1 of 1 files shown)


ojura:
Great addition, you beat me to it :-) Could you just prefix the string `Trajectory ` in front so it is more clear in the UI what the checkbox means?

jihoonl:
ping. is there any blocking issue in this pr?

ojura:
Since no one has been assigned to the PR, it might be a good idea to use mentions for pinging (e.g. @wohe).

wohe:
Sorry, this week most of the team was out of office, it was only blocked by lack of time.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -175,6 +175,7 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodeList() {
        ++trajectory_id) {
     const auto& single_trajectory_nodes = all_trajectory_nodes[trajectory_id];
     visualization_msgs::Marker marker;
+    marker.ns = "Trajectory " + std::to_string(trajectory_id);
     marker.id = marker_id++;
```
wohe: My understanding is that this only has to be unique in a namespace. Could we simplify now by removing `marker_id`, initializing to `0` here and doing a `++marked.id` below?

ojura: Marker id is not equal to trajectory id because we split the trajectory into multiple markers when there are more than 16384 trajectory nodes.

ojura: Sorry, I haven't read this carefully. Perhaps we could. You want to make `marker` a local variable of one scope above and to keep the marker id inside of it?

wohe: Yes, but I was arguing for the opposite: Every trajectory now has its own `ns`, so can have a markers with `id` 0, 1, 2, etc. Only the pair of `ns` and `id` has to be unique, see https://github.com/ros-visualization/rviz/blob/kinetic-devel/src/rviz/default_plugin/marker_display.h#L62

wohe: No, I want to start counting from 0 for each `ns`.

ojura: I see. I finished this off in another PR.

jihoonl:
Followed by #473

PR #454, opened 2017-07-27
Fix fixed frame issues of the RViz SubmapsDisplay.

wohe:
Adds a 'map_node_' which gets updated with the transform to RViz's
fixed frame. Before, submaps rendered incorrectly if a fixed frame
different from the map frame was chosen.

Removes the map frame property since this data is published on the
submap list topic.

Renames node names to reflect the frames they represent.

### DIFF
--- a/cartographer_rviz/cartographer_rviz/submaps_display.cc
+++ b/cartographer_rviz/cartographer_rviz/submaps_display.cc
@@ -49,9 +49,6 @@ SubmapsDisplay::SubmapsDisplay() : tf_listener_(tf_buffer_) {
   submap_query_service_property_ = new ::rviz::StringProperty(
       "Submap query service", kDefaultSubmapQueryServiceName,
       "Submap query service to connect to.", this, SLOT(Reset()));
-  map_frame_property_ = new ::rviz::StringProperty(
-      "Map frame", kDefaultMapFrame, "Map frame, used for fading out submaps.",
-      this);
   tracking_frame_property_ = new ::rviz::StringProperty(
       "Tracking frame", kDefaultTrackingFrame,
       "Tracking frame, used for fading out submaps.", this);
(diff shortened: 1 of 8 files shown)


cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -162,20 +165,31 @@ void SubmapsDisplay::processMessage(
...
-  try {
-    const ::geometry_msgs::TransformStamped transform_stamped =
-        tf_buffer_.lookupTransform(map_frame_property_->getStdString(),
-                                   tracking_frame_property_->getStdString(),
-                                   ros::Time(0) /* latest */);
-    for (auto& trajectory : trajectories_) {
-      for (auto& submap_entry : trajectory.second) {
-        submap_entry.second->SetAlpha(
-            transform_stamped.transform.translation.z);
+  if (map_frame_ != nullptr) {
```
ensonic: change to early exit?

cartographer_rviz/cartographer_rviz/submaps_display.cc
```diff
@@ -162,20 +165,31 @@ void SubmapsDisplay::processMessage(
...
+      map_node_->setPosition(position);
+      map_node_->setOrientation(orientation);
+      context_->queueRender();
+    }
+    // Update the fading by z distance.
+    try {
+      const ::geometry_msgs::TransformStamped transform_stamped =
+          tf_buffer_.lookupTransform(*map_frame_,
+                                     tracking_frame_property_->getStdString(),
+                                     ros::Time(0) /* latest */);
```
ensonic: would it be better to use a local var with ros::Time(0) to use the same timestamp here and above.

wohe:
PTAL

PR #455, opened 2017-07-28
 Add toggleable trajectories to RViz plugin

SirVer:


SirVer:
Alternative to #422.

SirVer:
Mhh, I cannot push to the original branch since this was not a PR by @ojura. Closing this and opening #456 with a branch from my fork instead.

PR #456, opened 2017-07-28
 Add toggleable trajectories to RViz plugin

SirVer:
Simpler implementation of #422 by @ojura. Original feature idea and implementation by @jihoonl.

### DIFF
--- a/cartographer_rviz/cartographer_rviz/submaps_display.h
+++ b/cartographer_rviz/cartographer_rviz/submaps_display.h
@@ -32,6 +32,21 @@
 
 namespace cartographer_rviz {
 
+// This should be a private class in SubmapsDisplay, unfortunately, QT does not
+// allow for this.
+struct Trajectory : public QObject {
+  Q_OBJECT
+
+ public:
+  Trajectory(std::unique_ptr<::rviz::BoolProperty> property);
+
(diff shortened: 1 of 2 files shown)


cartographer_rviz/cartographer_rviz/submaps_display.h
```diff
@@ -32,6 +32,21 @@
 
 namespace cartographer_rviz {
 
+// This should be a private class in SubmapsDisplay, unfortunately, QT does not
+// allow for this.
+struct Trajectory : public QObject {
+  Q_OBJECT
+
+ public:
+  Trajectory(std::unique_ptr<::rviz::BoolProperty> property);
```
ojura: @SirVer You can pass unique pointers by value in constructors? How does this exactly work?

macmason: Note the call to `std::move` further down. This pattern forces a move.

ojura: Yeah, that part didn't confuse me, I added it in the first place :-) Even if you accept an rvalue reference in your prototype, it actually becomes an lvalue, so you still need the `std::move` cast. What confused me is that when @SirVer removed the rvalue reference from the variable property, this introduced an additional implicit move from the rvalue at the call site into the property variable.

PR #461, opened 2017-07-31
Fix crash from memory corruption by switching unique_ptr for shared_ptr.

allenh1:
This fixes a new crash caused by the use of `unique_ptr`'s.

Without this, things crash immediately after the callback exits (when using IMU data).

### DIFF
--- a/cartographer_ros/cartographer_ros/sensor_bridge.cc
+++ b/cartographer_ros/cartographer_ros/sensor_bridge.cc
@@ -59,7 +59,7 @@ void SensorBridge::HandleOdometryMessage(
   }
 }
 
-std::unique_ptr<::cartographer::sensor::ImuData> SensorBridge::ToImuData(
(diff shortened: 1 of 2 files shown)


allenh1:
I signed it!

wohe:
Could you please explain why and what you believe this to fix? The code before looks correct to me.

allenh1:
@wohe the `unique_ptr` usage was consistently crashing cartographer_ros whenever I ran cartographer_ros. The `shared_ptr` fixes this crash.

I think the issue with the `unique_ptr` is the fact that the IMU message is not exclusive to the cartographer node, so a unique pointer was causing memory corruption.

wohe:
Where do you see multiple owners of the `unique_ptr`? Still looks good to me. That your crashes correlate with your change also does not convince me that this is causing/fixing it.

Assuming this is not the cause: Can you provide the console output from the crashes you see so that we can debug?

allenh1:
I must have caught things between merges yesterday. I can't seem to reproduce this. I'll report back if that changes.

PR #468, opened 2017-08-01
Keep offline node running, republish the clock

ojura:
Addresses #407 by @jihoonl, and #368 (partially). Previous attempt was in #414, but we concluded that it is not a good approach.

#443 made `Node` thread-safe, which enables us to launch an asynchronous spinner which will handle all ROS callbacks such as message publishing and servicing submap requests. Even when all bag processing ends, we want to keep publishing the submap list and answering submap queries (this is controlled with a new command-line argument in the offline node).

One small problem is that `Node::FinalizeTrajectory` is still a blocking function and keeps the mutex for quite a while, so you can not really view the final optimization in progress, but that we can handle that later.

### DIFF
--- a/cartographer_ros/cartographer_ros/offline_node_main.cc
+++ b/cartographer_ros/cartographer_ros/offline_node_main.cc
@@ -63,6 +63,7 @@ namespace {
 constexpr char kClockTopic[] = "clock";
 constexpr char kTfStaticTopic[] = "/tf_static";
 constexpr char kTfTopic[] = "tf";
+constexpr double kClockPublishFrequency = 30.;
 
 // TODO(hrapp): This is duplicated in node_main.cc. Pull out into a config
 // unit.
@@ -129,6 +130,7 @@ void Run(const std::vector<string>& bag_filenames) {
 
   ros::AsyncSpinner async_spinner(1, ros::getGlobalCallbackQueue());
   async_spinner.start();
+  rosgraph_msgs::Clock clock;
 
   for (const string& bag_filename : bag_filenames) {
     if (!::ros::ok()) {
@@ -197,7 +199,6 @@ void Run(const std::vector<string>& bag_filenames) {
               trajectory_id, topic,
               delayed_msg.instantiate<nav_msgs::Odometry>());
         }
-        rosgraph_msgs::Clock clock;
         clock.clock = delayed_msg.getTime();
         clock_publisher.publish(clock);
 
@@ -238,8 +239,10 @@ void Run(const std::vector<string>& bag_filenames) {
   node.SerializeState(bag_filenames.front() + ".pbstream");
 
   if (FLAGS_keep_running) {
-    ros::WallRate rate(20);
-    while (!::ros::ok()) {
+    ::ros::WallRate rate(kClockPublishFrequency);
+    while (::ros::ok()) {
+      clock_publisher.publish(clock);
+      ::ros::spinOnce();
       rate.sleep();
     }
   }
--- a/cartographer_ros/cartographer_ros/offline_node_main.cc
+++ b/cartographer_ros/cartographer_ros/offline_node_main.cc
@@ -53,6 +53,9 @@ DEFINE_bool(use_bag_transforms, true,
             "Whether to read, use and republish the transforms from the bag.");
 DEFINE_string(pbstream_filename, "",
               "If non-empty, filename of a pbstream to load.");
+DEFINE_bool(keep_running, false,
+            "Keep running the offline node after all messages from the bag "
+            "have been processed.");
 
 namespace cartographer_ros {
 namespace {
@@ -123,6 +126,9 @@ void Run(const std::vector<string>& bag_filenames) {
   if (urdf_transforms.size() > 0) {
     static_tf_broadcaster.sendTransform(urdf_transforms);
   }
+
+  ros::AsyncSpinner async_spinner(1, ros::getGlobalCallbackQueue());
+  async_spinner.start();
 
   for (const string& bag_filename : bag_filenames) {
     if (!::ros::ok()) {
@@ -195,8 +201,6 @@ void Run(const std::vector<string>& bag_filenames) {
         clock.clock = delayed_msg.getTime();
         clock_publisher.publish(clock);
 
-        ::ros::spinOnce();
-
         LOG_EVERY_N(INFO, 100000)
             << "Processed " << (delayed_msg.getTime() - begin_time).toSec()
             << " of " << duration_in_seconds << " bag time seconds...";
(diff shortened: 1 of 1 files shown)


ojura:
By the way, I think I haven't experienced a crash with Ctrl-C and ros::ok(). If this is the case, can we revert to the ROS interrupt handler?

wohe:
Can you split off the change to `ros::ok()` to a separate PR please, since it is unrelated?

Also, not to observing crashes is not enough to make this a good change. That said, I looked at #230 again in which I sadly did not include the crash it was intended to fix. There is the obvious problem before #230 that trajectories are not correctly finished which does cause crashes. This is fixed by the change of `return` to `break` which you retain. It does not need a signal handler. So I'm in favor of your suggested change.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -232,6 +237,15 @@ void Run(const std::vector<string>& bag_filenames) {
 #endif
 
   node.SerializeState(bag_filenames.front() + ".pbstream");
+
+  if (FLAGS_keep_running) {
+    ::ros::WallRate rate(kClockPublishFrequency);
+    while (::ros::ok()) {
+      clock_publisher.publish(clock);
+      ::ros::spinOnce();
```
ojura: Sorry, I messed up, you are right.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -124,6 +128,10 @@ void Run(const std::vector<string>& bag_filenames) {
     static_tf_broadcaster.sendTransform(urdf_transforms);
   }
 
+  ros::AsyncSpinner async_spinner(1, ros::getGlobalCallbackQueue());
```
ojura: Fixed, also added a constant which clarifies the number of threads for the `AsyncSpinner`.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -53,13 +53,17 @@ DEFINE_bool(use_bag_transforms, true,
...
+            "Keep running the offline node after all messages from the bag "
+            "have been processed.");
 
 namespace cartographer_ros {
 namespace {
 
 constexpr char kClockTopic[] = "clock";
 constexpr char kTfStaticTopic[] = "/tf_static";
 constexpr char kTfTopic[] = "tf";
+constexpr double kClockPublishFrequency = 30.;
```
ojura: Done.

ojura:
By the way, I have also made one more modification: the clock republishing functionality has been moved into a timer which is started right after a bag is finished (but before calling `FinalizeTrajectory`, which might take a while). The final loop now only waits for Ctrl-C.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -124,6 +129,16 @@ void Run(const std::vector<string>& bag_filenames) {
     static_tf_broadcaster.sendTransform(urdf_transforms);
   }
 
+  ros::AsyncSpinner async_spinner(kSingleThreaded);
+  async_spinner.start();
+  rosgraph_msgs::Clock clock;
+  auto clock_republish_timer = node.node_handle()->createWallTimer(
```
wohe: This seems a lot more complicated then what you had before. What is the advantage?

ojura: The advantage is that we don't have to wait for trajectory finalization to complete (which can take a while for larger bags) to start republishing the clock. I have a use case where I constantly need the clock published, without dropouts caused by the main thread having to wait for finalization.

wohe: Thanks for the explanation. Good for now, but in general let's try to keep code reviews focused and move further feature work to separate PRs.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -232,6 +247,12 @@ void Run(const std::vector<string>& bag_filenames) {
 #endif
 
   node.SerializeState(bag_filenames.front() + ".pbstream");
+  if (FLAGS_keep_running) {
+    ::ros::WallRate rate(10.);
```
wohe: Use `ros::waitForShutdown()`?

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -137,6 +152,8 @@ void Run(const std::vector<string>& bag_filenames) {
     rosbag::View view(bag);
     const ::ros::Time begin_time = view.getBeginTime();
     const double duration_in_seconds = (view.getEndTime() - begin_time).toSec();
+    // The message processing loop publishes the clock.
```
wohe: I think it is more readable if you move this down after `FinishTrajectory()` and restart before waiting for shutdown. Then you can also change the comment, i.e. say that we publish while waiting for trajectory finalization.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -124,6 +129,16 @@ void Run(const std::vector<string>& bag_filenames) {
     static_tf_broadcaster.sendTransform(urdf_transforms);
   }
 
+  ros::AsyncSpinner async_spinner(kSingleThreaded);
+  async_spinner.start();
+  rosgraph_msgs::Clock clock;
+  auto clock_republish_timer = node.node_handle()->createWallTimer(
+      ::ros::WallDuration(1. / kClockPublishFrequencyHz),
```
wohe: Consistency: Use a `kClockPublishPeriodSec` instead of a frequency.

cartographer_ros/cartographer_ros/offline_node_main.cc
ojura: Nice. I had been trying to find if something like this exists, but my Google-fu failed me.

cartographer_ros/cartographer_ros/offline_node_main.cc
ojura: Sure. Can I put it before serialization? That might take some time as well.

ojura: On the other hand, it might be easier if I leave it like it is now, so I don't have to worry about the code between timer.stop() and timer.start() such as serialization taking too long. Or can I put a timer.start right after the bag processing loop?

wohe: No. Please make it more readable. Putting a `start()` call before `SerializeState()` is fine. In general it is a good idea to write your code like you explain it to a human ("readability"). You explain your code as publishing during `FinishTrajectory()`, `SerializeState()` and waiting for shutdown, so unless there is a good reason to deviate from this, please implement it in the same way.

ojura: > publishing during FinishTrajectory(), SerializeState() and waiting for shutdown

I wanted to keep it this way because I found it more readable to describe as "republish clock always, except during message processing" rather than "publish during finish trajectory finalization, state serialization and waiting for shutdown (and eventually during something else we might add at the end)". I initially mentioned to you only trajectory finalization as being problematic, but what I had in mind was more along these lines. However, I can add it before SerializeState(), if you think it's more readable, no problem. :)

ojura:
Okay, I believe it should be ready now.

PR #470, opened 2017-08-03
Remove custom SIGINT handler from offline node

ojura:
I have singled here out the first commit from the other pull request.

### DIFF
--- a/cartographer_ros/cartographer_ros/offline_node_main.cc
+++ b/cartographer_ros/cartographer_ros/offline_node_main.cc
@@ -16,7 +16,6 @@
 
 #include <time.h>
 #include <chrono>
-#include <csignal>
 #include <sstream>
 #include <string>
 #include <vector>
--- a/cartographer_ros/cartographer_ros/offline_node_main.cc
+++ b/cartographer_ros/cartographer_ros/offline_node_main.cc
@@ -62,10 +62,6 @@ constexpr char kClockTopic[] = "clock";
 constexpr char kTfStaticTopic[] = "/tf_static";
 constexpr char kTfTopic[] = "tf";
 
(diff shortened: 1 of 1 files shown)


wohe:
Copying the rationale from #468: There is the obvious problem before #230 that trajectories are not correctly finished which does cause crashes. This is fixed by the change of `return` to `break` which you retain. It does not need a signal handler.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -252,9 +248,7 @@ int main(int argc, char** argv) {
       << "-configuration_basename is missing.";
   CHECK(!FLAGS_bag_filenames.empty()) << "-bag_filenames is missing.";
 
-  std::signal(SIGINT, &::cartographer_ros::SigintHandler);
-  ::ros::init(argc, argv, "cartographer_offline_node",
-              ::ros::init_options::NoSigintHandler);
+  ::ros::init(argc, argv, "cartographer_offline_node");
```
wohe: I think the `#include <csignal>` can also be removed.

ojura: Updated

ojura:
> trajectories are not correctly finished which does cause crashes

Sounds like `Node` should check in its destructor if there are any unfinalized trajectories.

wohe:
That's what I meant, it is `CHECK`ed. In the `OrderedMultiQueue` though: https://github.com/googlecartographer/cartographer/blob/97bb74157f14267c43f6b0c44eda19e7ed6336e1/cartographer/sensor/ordered_multi_queue.cc#L45

PR #471, opened 2017-08-03
Deduplicate LoadOptions().

wohe:


cartographer_ros/cartographer_ros/trajectory_options.h
```diff
@@ -21,7 +21,7 @@
 
 #include "cartographer/common/lua_parameter_dictionary.h"
 #include "cartographer/common/port.h"
-#include "cartographer_ros/sensor_bridge.h"
+#include "cartographer/mapping/proto/trajectory_builder_options.pb.h"
```
ensonic: Is this and the change above related?

wohe: Yes. This header is now needed by `node_options.h`, but it before included `sensor_bridge.h` which is not needed. We remove it to avoid additional unnecessary transitive includes where `node_options.h` get included.

PR #473, opened 2017-08-04
Use trajectory id as namespace for trajectory marker

ojura:
Continuation of #452.


damonkohler:
@jihoonl I assume this is OK with you since you closed #452 and referred to this one.

jihoonl:
yeah no problem.

PR #479, opened 2017-08-08
Also extrapolate based on odometry if available.

wohe:


cartographer_ros/cartographer_ros/node.cc
```diff
@@ -448,6 +448,11 @@ void Node::HandleOdometryMessage(const int trajectory_id,
                                  const string& sensor_id,
                                  const nav_msgs::Odometry::ConstPtr& msg) {
   carto::common::MutexLocker lock(&mutex_);
+  auto sensor_bridge_ptr = map_builder_bridge_.sensor_bridge(trajectory_id);
```
damonkohler: Either inline or also use on line 456. Also, these pointers should be const, no?

wohe: Thanks, I somehow forgot.

BrannonKing:
Can you explain this change further? I would like to understand the necessity of merging it into my branch. I see it happened several weeks back for the IMU data.

What is the new position extrapolator? 
Surely we were estimating position previously?
Can we use the extrapolator to get a better position or position variance?
Does the extrapolator need to be incorporated into the laser scan data as well?

wohe:
Sure.

> What is the new position extrapolator?

Pose extrapolator you mean? See https://github.com/googlecartographer/cartographer/blob/master/cartographer/mapping/pose_extrapolator.h. It can extrapolate rotations from poses if no IMU is available and can combine linear velocity from odometry with rotation from IMU.

> Surely we were estimating position previously?

We estimate positions in many places. Can you be more specific? Until recently, if tf was published at high frequency, it repeated the pose of the last scan match. This also uses the odometry data for higher frequency tf.

> Can we use the extrapolator to get a better position or position variance?

To compute a variance? No. To estimate a pose with smaller error? Maybe.

> Does the extrapolator need to be incorporated into the laser scan data as well?

Right now it is only used in Cartographer ROS for tf. I plan to use it in the local trajectory builders for 2D and 3D as well. It will then influence the unwarping of laser scans. Does that answer your question?

> Is the code in a runnable/workable state without this change?

There can always be bugs, but other than that: Of course. We aim to never merge a PR that leaves the code in a bad state.

BrannonKing:
Thank you for the fast and thorough response. 
> Does that answer your question?

Yes. I'm satisfied with your responses.
> We estimate positions in many places. Can you be more specific? 

I was referring to the transform broadcast from `Node::PublishTrajectoryStates`.
> It can extrapolate rotations from poses if no IMU is available and can combine linear velocity from odometry with rotation from IMU.

That's interesting that you suggest that it can use linear velocities. I would like to input those directly. For my particular use case, I have body linear and angular velocities available in the tracking frame (from a wheel encoder source combo). I also have nice IMU data available, but it's not in the tracking frame. Cartographer traditionally rejects my IMU data because of its need for translation (which message, BTW, seems incorrect as accelerations can be translated using this dynamics formula for centrifugal force deltas:  ω×(ω×t) +ω'×t ). I would like to use the angular and linear velocities in addition to the IMU. Right now there is no way to directly input linear velocities. I don't see a way to add them directly to the estimator interface.

If we're using the estimator, do we still need to send the IMU data into the trajectory builder? Why does the data need to go to both locations? Or it won't once the estimator's pose is going into the trajectory builder, which piece is still missing?

wohe:
> I was referring to the transform broadcast from Node::PublishTrajectoryStates.

This is the place I was describing. We recently added extrapolation and are now also making use of odometry.

> That's interesting that you suggest that it can use linear velocities. I would like to input those directly.

We only use odometry poses as input and compute velocities from there. I currently think that this is just a small inconvenience if you have velocities already. Since we want to make use of the odometry poses more in the future, I'd prefer to keep the data in that form.

> Cartographer traditionally rejects my IMU data because of its need for translation (which message, BTW, seems incorrect as accelerations can be translated using this dynamics formula for centrifugal force deltas: ω×(ω×t) +ω'×t ). 

This message should be read as a suggestion to change the tracking frame, not stop you from using your IMU data. I still think the message is correct: I'm not familiar with your formula in particular, but it contains ω' which I assume is angular accelerations. This is exactly the inaccuracy we talk about: angular accelerations are not measured by an IMU, but contribute to linear acceleration as seen from a coordinate frame with a different origin. Since there is nothing special about the tracking frame, colocating it with the IMU frame is unproblematic and resolves this issue.

> If we're using the estimator, do we still need to send the IMU data into the trajectory builder? Why does the data need to go to both locations? Or it won't once the estimator's pose is going into the trajectory builder, which piece is still missing?

Yes. It's used in both. No.

PR #480, opened 2017-08-08
Release mutex in Node during final optimization

### LINKED ISSUE #476: Node::FinalizeTrajectory blocks callback servicing
As I mentioned in #468, one final issue is that `MapBuilderBridge::FinalizeTrajectory()` (more precisely, when `SparsePoseGraph::RunFinalOptimization()` calls `SparsePoseGraph::WaitForAllComputations()`) can take a long time and make `Node::FinishTrajectory()` hold the `Node` mutex for too long. This blocks servicing submap requests and all other types of callbacks.  @wohe do you have a suggestion how to do more fine-grained locking?

ojura:
Enables servicing callbacks during the final optimization. This (along with #443 and #468) enables watching in RViz the progress of the final optimization using the offline node.

One question: should the online node also call `RunFinalOptimization()` in `Node::HandleFinishTrajectory`, or only at the end (which is handled by `FinishAllTrajectories`) ? If so, could this be done asynchronously for similar reasons (to avoid stopping callback servicing)?

> No new data should arrive then as this is an assumption WaitForAllComputations() makes.

This is ensured by removing the sensor bridge in `MapBuilderBridge::FinishTrajectory`. Although, if we wanted to insert a new constraint (e.g. landmark), possibly even during the final optimization, this approach will have to be adapted, right?

Fixes #476.

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -77,8 +77,12 @@ void MapBuilderBridge::FinishTrajectory(const int trajectory_id) {
   // Make sure there is a trajectory with 'trajectory_id'.
   CHECK_EQ(sensor_bridges_.count(trajectory_id), 1);
   map_builder_.FinishTrajectory(trajectory_id);
-  map_builder_.sparse_pose_graph()->RunFinalOptimization();
   sensor_bridges_.erase(trajectory_id);
+}
+
+void MapBuilderBridge::RunFinalOptimization() {
+  LOG(INFO) << "Running final trajectory optimization...";
+  map_builder_.sparse_pose_graph()->RunFinalOptimization();
 }
 
 void MapBuilderBridge::SerializeState(const std::string& filename) {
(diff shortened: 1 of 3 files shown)


ojura:
Before merging, please check if this is what you want with respect to the trajectory finalization service (as I explained above - I did not add a call to `RunFinalOptimization` in `HandleFinishTrajectory`. If one should be added, I suggest it be done asynchronously.

ojura:
Damn, you beat me to it :-)

damonkohler:
I think this is the right way to do things. You've separated the concept of running the final optimization from finishing trajectories. As you say, the final optimization is run when calling "FinishAllTrajectories." That makes sense. After all, it is a "final" optimization and can only be "final" if it's called once all trajectories are finished.

ojura:
I see, that makes sense. Although, it now occured to me that there's an asymmetry between the offline and the online node - the offline node still calls the final optimization at the end of each trajectory when it is finished before it goes on to process the next bag (trajectory).

damonkohler:
SGTM

PR #482, opened 2017-08-09
Feature/add flag to start first trajectory

yutakaoka:
@damonkohler I added a flag to disable _StartTrajectoryWithDefaultTopics_. Please review it.

### DIFF
--- a/cartographer_ros/cartographer_ros/node_main.cc
+++ b/cartographer_ros/cartographer_ros/node_main.cc
@@ -28,6 +28,9 @@ DEFINE_string(configuration_basename, "",
               "Basename, i.e. not containing any directory prefix, of the "
               "configuration file.");
 DEFINE_string(map_filename, "", "If non-empty, filename of a map to load.");
+DEFINE_bool(start_trajectory_with_default_topics, true,
+            "Enable/disable to start the first trajectory with default "
+            "topics.");
 
 namespace cartographer_ros {
 namespace {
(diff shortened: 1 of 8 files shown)


cartographer_ros/cartographer_ros/node_main.cc
```diff
@@ -38,14 +41,20 @@ void Run() {
...
   NodeOptions node_options;
   TrajectoryOptions trajectory_options;
-  std::tie(node_options, trajectory_options) =
-      LoadOptions(FLAGS_configuration_directory, FLAGS_configuration_basename);
-
+  if (FLAGS_start_trajectory_with_default_topics) {
+    std::tie(node_options, trajectory_options) = LoadOptions(
+        FLAGS_configuration_directory, FLAGS_configuration_basename);
+  } else {
+    node_options = LoadNodeOptions(FLAGS_configuration_directory,
```
damonkohler: It's not clear to me why you took this approach. Why not simply only add the if statement on line 55?

cartographer_ros/cartographer_ros/node_main.cc
```diff
@@ -28,6 +28,9 @@ DEFINE_string(configuration_basename, "",
               "Basename, i.e. not containing any directory prefix, of the "
               "configuration file.");
 DEFINE_string(map_filename, "", "If non-empty, filename of a map to load.");
+DEFINE_bool(
+    start_trajectory_with_default_topics, true,
+    "Enable/disable to start the first trajectory with default topics.");
```
damonkohler: For consistency: "Enable to immediately start the first trajectory with default topics."

cartographer_ros/cartographer_ros/node_main.cc
yutakaoka: I think we don't need _trajectory_options_ if _start_trajectory_with_default_topics_ is false. Is that correct? Or should we fill blank for trajectory_options topics on lua file?

damonkohler: I see, that makes sense.

In that case, it seems to me that LoadOptions and LoadNodeOptions should be moved into node_main.cc to remove the duplication of loading the LuaParameterDictionary in LoadOptions and LoadNodeOptions. It would also remove the confusing naming of those methods.

You would simply load the Lua configuration, then call CreateNodeOptions and CreateTrajectoryOptions as appropriate all directly in node_main.cc.

wohe: Please do not move this into `node_main.cc` and have a look at other uses first. Loading options is also done in the offline node and in the configuration file tests in this and other Cartographer repos.

yutakaoka: I saw another _LoadOptions()_ in `start_trajectory_main.cc`. It seems like returning TrajectoryOptions. I think we may want to clean up around LoadOptions later? For now, I am thinking of just adding the if statement on _StartTrajectoryWithDefaultTopics_(line 55). How does it sound?

wohe: Sounds good to me.

cartographer_ros/cartographer_ros/node_main.cc
yutakaoka: Done

cartographer_ros/cartographer_ros/node_main.cc
yutakaoka: Done

yutakaoka:
ping @damonkohler , How does it sound about this PR?

PR #483, opened 2017-08-10
Perform final optimization at the end

ojura:
@damonkohler here's a small follow-up to #480 based on what you said there. There is now a clear distinction between finishing the trajectory and running the final optimization in `Node`. The offline node marks the trajectories finished as soon as it reaches the end of the respective bag, while the final optimization is run at the very end.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -428,26 +428,32 @@ bool Node::HandleWriteState(
 }
 
 void Node::FinishAllTrajectories() {
-  {
-    carto::common::MutexLocker lock(&mutex_);
-    for (auto& entry : is_active_trajectory_) {
-      const int trajectory_id = entry.first;
-      if (entry.second) {
-        map_builder_bridge_.FinishTrajectory(trajectory_id);
-        entry.second = false;
-      }
+  carto::common::MutexLocker lock(&mutex_);
+  for (auto& entry : is_active_trajectory_) {
+    const int trajectory_id = entry.first;
+    if (entry.second) {
+      map_builder_bridge_.FinishTrajectory(trajectory_id);
+      entry.second = false;
     }
   }
-  map_builder_bridge_.RunFinalOptimization();
 }
 
 void Node::FinishTrajectory(const int trajectory_id) {
+  carto::common::MutexLocker lock(&mutex_);
+  CHECK(is_active_trajectory_.at(trajectory_id));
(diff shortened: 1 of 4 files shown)


cartographer_ros/cartographer_ros/node.cc
```diff
@@ -428,26 +427,35 @@ bool Node::HandleWriteState(
...
   }
-  map_builder_bridge_.RunFinalOptimization();
 }
 
 void Node::FinishTrajectory(const int trajectory_id) {
+  carto::common::MutexLocker lock(&mutex_);
+  FinishTrajectoryUnderLock(trajectory_id);
+}
+
+void Node::FinishTrajectoryUnderLock(const int trajectory_id) {
```
wohe: Remove this convenience function? It does not seem to improve readability.

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -209,16 +209,15 @@ void Run(const std::vector<string>& bag_filenames) {
...
-    // which might take a while.
-    clock_republish_timer.start();
     node.FinishTrajectory(trajectory_id);
-    clock_republish_timer.stop();
   }
 
-  // Republish the clock after bag processing has been completed.
+  // Ensure the clock is republished during the final optimization, which might
+  // take a while.
   clock_republish_timer.start();
```
wohe: Remove empty line?

cartographer_ros/cartographer_ros/offline_node_main.cc
```diff
@@ -209,16 +209,15 @@ void Run(const std::vector<string>& bag_filenames) {
...
     bag.close();
-    // Ensure the clock is republished also during trajectory finalization,
-    // which might take a while.
-    clock_republish_timer.start();
     node.FinishTrajectory(trajectory_id);
-    clock_republish_timer.stop();
   }
 
-  // Republish the clock after bag processing has been completed.
+  // Ensure the clock is republished during the final optimization, which might
```
wohe: This is not the only purpose, right? What about serialization and the `--keep_running` flag?

cartographer_ros/cartographer_ros/node.cc
ojura: Sure, I can revert the last commit. I went by the rule "three or more, use a helper function". :-)

ojura: I have reverted the commit which introduced the helper function.

PR #499, opened 2017-09-08
Fixed ratio sensor sampling in cartographer_ros

ojura:
@cschuet here's an alternate implementation of https://github.com/googlecartographer/cartographer/pull/502 in cartographer_ros.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.h
+++ b/cartographer_ros/cartographer_ros/node.h
@@ -23,6 +23,7 @@
 #include <unordered_set>
 #include <vector>
 
+#include "cartographer/common/fixed_ratio_sampler.h"
 #include "cartographer/common/mutex.h"
 #include "cartographer/mapping/pose_extrapolator.h"
 #include "cartographer_ros/map_builder_bridge.h"
@@ -115,6 +116,7 @@ class Node {
                          int trajectory_id);
   void PublishSubmapList(const ::ros::WallTimerEvent& timer_event);
   void AddExtrapolator(int trajectory_id, const TrajectoryOptions& options);
+  void AddSensorSamplers(int trajectory_id, const TrajectoryOptions& options);
   void PublishTrajectoryStates(const ::ros::WallTimerEvent& timer_event);
   void PublishTrajectoryNodeList(const ::ros::WallTimerEvent& timer_event);
   void PublishConstraintList(const ::ros::WallTimerEvent& timer_event);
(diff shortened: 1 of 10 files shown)


cartographer_ros/cartographer_ros/node.h
```diff
@@ -138,8 +140,22 @@ class Node {
...
+          imu_sampler(imu_sampling_ratio) {}
+
+    ::cartographer::common::FixedRatioSampler rangefinder_sampler;
+    ::cartographer::common::FixedRatioSampler odometry_sampler;
+    ::cartographer::common::FixedRatioSampler imu_sampler;
+  };
+
   // These are keyed with 'trajectory_id'.
   std::map<int, ::cartographer::mapping::PoseExtrapolator> extrapolators_;
+  std::map<int, TrajectorySensorSamplers> sensor_samplers_;
```
ojura: @cschuet do you know whether `map` or `unordered_map` should be used here?

cartographer_ros/cartographer_ros/trajectory_options.cc
```diff
@@ -59,6 +59,12 @@ TrajectoryOptions CreateTrajectoryOptions(
           "num_subdivisions_per_laser_scan");
   options.num_point_clouds =
       lua_parameter_dictionary->GetNonNegativeInt("num_point_clouds");
+  options.rangefinder_sampling_ratio =
+      lua_parameter_dictionary->GetDouble("rangefinder_sampling_ratio");
+  options.odometry_sampling_ratio =
+      lua_parameter_dictionary->GetDouble("odometry_sampling_ratio");
+  options.imu_sampling_ratio =
+      lua_parameter_dictionary->GetDouble("imu_sampling_ratio");
   CheckTrajectoryOptions(options);
```
ojura: should I check  0 <= ratio <= 1 here, or would you prefer if we introduced a [single check in `FixedRatioSampler`](https://github.com/googlecartographer/cartographer/pull/502/files#diff-eb647468b6eb0fcffce71020e654e168R24)?

cartographer_ros/configuration_files/backpack_2d.lua
```diff
@@ -32,6 +32,9 @@ options = {
   submap_publish_period_sec = 0.3,
   pose_publish_period_sec = 5e-3,
   trajectory_publish_period_sec = 30e-3,
+  rangefinder_sampling_ratio = 1.,
```
ojura: Keep in mind that you will have to update all other configuration files (for example, in repos `cartographer_turtlebot`, `cartographer_toyota_hsr`, etc.). This was one reason why I was reluctant to introduce a new option to cartographer_ros. 

Perhaps it might make sense to introduce a common .lua file in cartographer_ros which would have some default settings, and which would be included from other robot-specific .lua files. This way, when a new option is introduced, a default value could be provided in the common settings file, without having to update all other configurations.

cartographer_ros/cartographer_ros/node.h
ojura: @cschuet I decided to use an unordered map, since we're not iterating through this anywhere. Maybe you can resolve the other questions?

cartographer_ros/configuration_files/backpack_2d.lua
wohe: @cschuet Please make sure this is handled when you merge.

cartographer_ros/cartographer_ros/trajectory_options.cc
ojura: @wohe no need to check if the ratio is sane?

wohe: If we add a `CHECK`, we should do it in `FixedRatioSampler()` to avoid the duplication. Doesn't seem like a big issue to me though. Right now we just interpret values clamped to [0, 1].

ojura: Okay, I have edited the pull request in googlecartographer/cartographer#502, which initially contained the sensor sampling and ratio checking, to retain only the ratio check in `FixedRatioSampler()`.

PR #500, opened 2017-09-08
disconnects trajectory if the middle trajectory nodes are trimmed

jihoonl:
Related: 
* https://github.com/googlecartographer/cartographer/issues/409
* https://github.com/googlecartographer/cartographer/pull/472
* https://github.com/googlecartographer/cartographer/pull/422


cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -36,6 +36,22 @@ ::std_msgs::ColorRGBA ToMessage(const cartographer::io::FloatColor& color) {
   return result;
 }
 
+visualization_msgs::Marker CreateTrajectoryMarker(int trajectory_id, int split,
```
wohe: nit: `const int` in the implementation.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -36,6 +36,22 @@ ::std_msgs::ColorRGBA ToMessage(const cartographer::io::FloatColor& color) {
...
+                                                  const std::string& frame_id) {
+  visualization_msgs::Marker marker;
+  marker.ns = "Trajectory " + std::to_string(trajectory_id) + "." +
+              std::to_string(split);
```
wohe: Having a separate namespace for each split would lead to many namespaces shown in RViz. Also, the labels will change when the trajectory is split. Any reason why we don't use one namespace per trajectory and increase the `marker.id` only?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -177,20 +193,23 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodeList() {
...
-    marker.pose.orientation.w = 1.0;
-    marker.pose.position.z = 0.05;
+    bool jump = false;
+    int num_splits = 0;
+    auto marker = CreateTrajectoryMarker(trajectory_id, num_splits,
+                                         node_options_.map_frame);
+
     for (const auto& node : single_trajectory_nodes) {
       if (node.trimmed()) {
+        jump = true;
```
wohe: Similar to the handling large markers: Push the marker here if it contains a line and always reset it? Then we do not need the `jump` variable.

jihoonl: `jump` is necessary. It would push empty markers as many as trimmed nodes if we push marker here.

wohe: That's what I meant by "if it contains a line", i.e. at least two points. Maybe even extract a function to do the `PushAndResetLineMarker()` and use it in all three places.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
jihoonl: It was a small indication to show how many splits happened in that trajectory. And I think it can only grow only to num_submaps in the trajectory. I can revert it back if you don't like this. Or I can append node_index instead of num split. How about this?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
jihoonl: ah yes you are right.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
wohe: If we remove single nodes there could be more. In what case would you find providing this information in RViz useful? If we keep it one trajectory ID per namespace, hiding everything related to a trajectory ID is much easier which seems a good thing to me.

jihoonl: I agree that it will be too many if cartographer removes single nodes. I am fine with having one namespace per trajectory. It was just for convenient debuggability.

jihoonl:
done.

jihoonl:
Done.

jihoonl:
ping.

wohe:
CI passed at #544

PR #502, opened 2017-09-12
Serialize state before shutdown

jihoonl:
provide a new configuration to save the last state of cartographer before shutdown the node.


jihoonl:
ping.

jihoonl:
Man. I want to go there too. Have fun!

PR #504, opened 2017-09-12
configurable occupancy grid publishing speed

jihoonl:
introducing `publish_period_sec` to control the frequency of occupancy grid publisher. It is not necessary to publish occupancy grid for every submap updates.


SirVer:
lgtm. I changed some whitespace and merged master, so this can be merged as soon as CI is happy.

Thanks for the change!

PR #506, opened 2017-09-13
Allow multiple textures in ROS message.

gaschler:
Forward all textures from protobuf to ROS messages.

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -108,15 +108,18 @@ bool MapBuilderBridge::HandleSubmapQuery(
   CHECK(response_proto.textures_size() > 0)
       << "empty textures given for submap: " << submap_id;
 
-  // TODO(gaschler): Forward all textures, not just the first one.
-  const auto& texture_proto = *response_proto.textures().begin();
-  response.cells.insert(response.cells.begin(), texture_proto.cells().begin(),
-                        texture_proto.cells().end());
-  response.width = texture_proto.width();
-  response.height = texture_proto.height();
-  response.resolution = texture_proto.resolution();
-  response.slice_pose = ToGeometryMsgPose(
-      cartographer::transform::ToRigid3(texture_proto.slice_pose()));
+  response.submap_version = response_proto.submap_version();
+  for (const auto& texture_proto : response_proto.textures()) {
+    response.textures.emplace_back();
+    auto& texture = response.textures.back();
+    texture.cells.insert(texture.cells.begin(), texture_proto.cells().begin(),
+                         texture_proto.cells().end());
+    texture.width = texture_proto.width();
+    texture.height = texture_proto.height();
+    texture.resolution = texture_proto.resolution();
+    texture.slice_pose = ToGeometryMsgPose(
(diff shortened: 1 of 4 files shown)


cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -108,15 +108,18 @@ bool MapBuilderBridge::HandleSubmapQuery(
...
-  // TODO(gaschler): Forward all textures, not just the first one.
-  const auto& texture_proto = *response_proto.textures().begin();
-  response.cells.insert(response.cells.begin(), texture_proto.cells().begin(),
-                        texture_proto.cells().end());
-  response.width = texture_proto.width();
-  response.height = texture_proto.height();
-  response.resolution = texture_proto.resolution();
-  response.slice_pose = ToGeometryMsgPose(
-      cartographer::transform::ToRigid3(texture_proto.slice_pose()));
+  response.submap_version = response_proto.submap_version();
```
wohe: No need to set this twice? Remove!

cartographer_ros/cartographer_ros/submap.cc
```diff
@@ -30,29 +30,29 @@ std::unique_ptr<SubmapTexture> FetchSubmapTexture(
   ::cartographer_ros_msgs::SubmapQuery srv;
   srv.request.trajectory_id = submap_id.trajectory_id;
   srv.request.submap_index = submap_id.submap_index;
-  if (!client->call(srv)) {
+  if (!client->call(srv) || srv.response.textures.empty()) {
```
wohe: We don't want to send empty texture lists, so this would be a bug. Similar to before: Make this a `CHECK` instead of failing silently?

cartographer_ros/cartographer_ros/map_builder_bridge.cc
gaschler: Done

cartographer_ros/cartographer_ros/submap.cc
gaschler: Done

gaschler:
@wohe Please take a look at the changes.

cartographer_ros/cartographer_ros/submap.cc
```diff
@@ -30,29 +30,28 @@ std::unique_ptr<SubmapTexture> FetchSubmapTexture(
   ::cartographer_ros_msgs::SubmapQuery srv;
   srv.request.trajectory_id = submap_id.trajectory_id;
   srv.request.submap_index = submap_id.submap_index;
-  if (!client->call(srv)) {
-    return nullptr;
-  }
-  std::string compressed_cells(srv.response.cells.begin(),
-                               srv.response.cells.end());
+  CHECK(client->call(srv)) << "SubmapQuery call failed.";
```
SirVer: This is out of the control of the caller, i.e. an expected failure. Do not CHECK here.

cartographer_ros/cartographer_ros/submap.cc
```diff
@@ -30,29 +30,28 @@ std::unique_ptr<SubmapTexture> FetchSubmapTexture(
...
+  CHECK(!srv.response.textures.empty()) << "Response contains no textures.";
```
SirVer: remove text, it is tautologic with the CHECK.

Issue #508, opened 2017-09-19
Question about slice_pose !! 


GMahmoud:
Hey guys,

I am trying to build the cartographer submaps with unity3D. But I am a little bit confused about the slice_pose and its frame. I am not sure that the pose is expressed in the map frame. 
So could you please give me some help on that and tell me which frame (position and orientation) are you using to express the slice_pose ?  

Cheers

wohe:
`slice_pose` is the pose of the textured rectangle which is used for visualizing a submap in the submap frame. To get the pose of the textured rectangle in the map frame, the submap pose must be multiplied by it.

GMahmoud:
Thank you @wohe for your answer. I got this by adding some rviz axis to the ogre node.

Cheers

PR #510, opened 2017-09-25
OccupancyGrid(pgm and meta) exporter from pbstream

### LINKED ISSUE #475: Generating pgm/yaml map
Since the output of the 'finish_trajectory' service has changed, I'm unable to generate .pgm/.yaml map files for other nodes (non-Cartographer) to use.

I'm aware of the 'write_state' service which provides a .pbstream file, but is it still possible to write to a .pgm/.yaml file?

jihoonl:
This file loads submaps from pbstream file, generates occupancy grid, and exports map server compatible pgm image and yaml file.

Note that `WriteOccupancyGridToPgm` isn't very optimized. So, any optimization suggestion is welcome. And `WriteOccupancyGridInfoToYaml` is copy from old code.

resolves #475

### DIFF
--- a/cartographer_ros/CMakeLists.txt
+++ b/cartographer_ros/CMakeLists.txt
@@ -44,6 +44,7 @@ google_enable_testing()
 find_package(catkin REQUIRED COMPONENTS ${PACKAGE_DEPENDENCIES})
 
 include(FindPkgConfig)
+pkg_search_module(YAMLCPP REQUIRED yaml-cpp>=0.5.1)
 
 find_package(LuaGoogle REQUIRED)
 find_package(PCL REQUIRED COMPONENTS common io)
@@ -79,6 +80,7 @@ catkin_package(
     # TODO(damonkohler): This should be here but causes Catkin to abort because
     # protobuf specifies a library '-lpthread' instead of just 'pthread'.
     # CARTOGRAPHER
+    YAMLCPP
     PCL
     EIGEN3
     Boost
@@ -96,6 +98,10 @@ add_library(${PROJECT_NAME} ${ALL_SRCS})
 add_subdirectory("cartographer_ros")
 
 target_link_libraries(${PROJECT_NAME} PUBLIC cartographer)
+
+# YAML
+target_include_directories(${PROJECT_NAME} SYSTEM PUBLIC ${YAMLCPP_INCLUDE_DIRS})
+target_link_libraries(${PROJECT_NAME} PUBLIC ${YAMLCPP_LIBRARIES})
 
 # Lua
 target_include_directories(${PROJECT_NAME} SYSTEM PUBLIC ${LUA_INCLUDE_DIR})
--- a/cartographer_ros/cartographer_ros/CMakeLists.txt
+++ b/cartographer_ros/cartographer_ros/CMakeLists.txt
@@ -66,3 +66,14 @@ install(TARGETS cartographer_occupancy_grid_node
   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
+
+google_binary(cartographer_occupancy_grid_exporter
+  SRCS
+    occupancy_grid_exporter.cc
+)
+
+install(TARGETS cartographer_occupancy_grid_exporter
+  ARCHIVE DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
+  RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
+)
--- a/cartographer_ros/cartographer_ros/occupancy_grid.cc
+++ b/cartographer_ros/cartographer_ros/occupancy_grid.cc
@@ -163,19 +163,5 @@ void WriteOccupancyGridInfoToYaml(const OccupancyGridState& grid_state,
   yaml_file.close();
   CHECK(yaml_file) << "Writing '" << yaml_filename << "' failed.";
 }
-}
 
-/*
-for (size_t y = 0; y < gridnfo.height; ++y) {
-  for (size_t x = 0; x < grid.info.width; ++x) {
-    const size_t i = x + (grid.info.height - y - 1) * grid.info.width;
-    if (grid.data[i] >= 0 && grid.data[i] <= 100) {
-      pgm_file.put((100 - grid.data[i]) * 255 / 100);
-    } else {
-      // We choose a value between the free and occupied threshold.
-      constexpr uint8_t kUnknownValue = 128;
-      pgm_file.put(kUnknownValue);
-    }
-  }
-}
-*/
+}  // namespace cartographer_ros
--- a/cartographer_ros/cartographer_ros/occupancy_grid.cc
+++ b/cartographer_ros/cartographer_ros/occupancy_grid.cc
@@ -1,6 +1,10 @@
 
 #include "cartographer_ros/occupancy_grid.h"
 
+#include <fstream>
+
+#include "yaml-cpp/yaml.h"
+
 Eigen::Affine3d ToEigen(const ::cartographer::transform::Rigid3d& rigid3) {
   return Eigen::Translation3d(rigid3.translation()) * rigid3.rotation();
 }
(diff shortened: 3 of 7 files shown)


SirVer:
Why did you not do this in the asset writer? This seems to be fundamentally the same as bringing up Cartographer with the map and then running `map_server`.

jihoonl:
> Why did you not do this in the asset writer?

Because asset writer requires rosbag. Using rosbag is not an option for us.

> Your solution seems to be fundamentally the same as bringing up Cartographer with the map and then running map_server.

Yes, you are right. It is fundamentally the same as bringing up run cartographer with the map, run map_server, and map_saver. I implemented this exporter because running `roscore`,  `cartographer`, `occupancy grid node`, `map server`, and `map saver` is too much overhead to extract a map from a file.

SirVer:
> Because asset writer requires rosbag. Using rosbag is not an option for us.

Are you saying that you using a different data format to serialize your data to disk? Or that you do not serialize your sensor data?

> Yes, you are right. It is fundamentally the same as bringing up run cartographer with the map, run map_server, and map_saver. I implemented this exporter because running roscore, cartographer, occupancy grid node, map server, and map saver is too much overhead to extract a map from a file.

What you describing can be expressed in a single launch file - is that not a workable solution.

Here are my thoughts: The proposed way for `cartographer_ros` is to use the asset writer to generate probability grids (currently blocked on #475, but we'll fix this probably today). My understanding is that you are unable to use the asset_writer because you do not fully embrace the ROS standards, but swapped out rosbag for a custom implementation. Because of this, the asset writer is unavailable to you and you built this workaround - which should not be used by anybody that uses rosbag. This raises the question why it should be maintained in `cartographer_ros`. If this reasoning is sound, I am uncomfortable with merging this.

Instead it seems like we should think how you can make use of the assets writer, probably by providing an interface that is not tied to rosbag. Wdyt?

jihoonl:
> Here are my thoughts: The proposed way for cartographer_ros is to use the asset writer to generate probability grids (currently blocked on #475, but we'll fix this probably today). My understanding is that you are unable to use the asset_writer because you do not fully embrace the ROS standards, but swapped out rosbag for a custom implementation. Because of this, the asset writer is unavailable to you and you built this workaround - which should not be used by anybody that uses rosbag. This raises the question why it should be maintained in cartographer_ros. If this reasoning is sound, I am uncomfortable with merging this.

I don't get why you got an impression that we are using the different format of serialized data. It is just a extracting a data from pbstream file that is already generated. And it is just because we let the robot move lively around to SLAM the environment and generate pbstream file without recording rosbag data.

And since the generated pbstream file already included all submap images and their poses, it was easy to generate a unified version just as `occupancy_grid_node`.

> Instead it seems like we should think how you can make use of the assets writer, probably by providing an interface that is not tied to rosbag. Wdyt?

It can be alternative if asset writer can run only with pbstream file. I don't like the fact that I have to maintain another inflexible configuration file though.(e.g I can't dynamically(or via cmd line) configure the exported map file location.)

ojura:
I would like to divert your attention to googlecartographer/cartographer/issues/508, which might be a bit relevant.

In case you do not want to save bags, but only the serialized .pbstream state - if inside the trajectory nodes you also had the full range data stored as before (which is the subject of the issue), you would not need the bag at all to write assets, the pbstream would be enough.

jihoonl:
@ojura that's something else actually. I am currently pretty happy with not having range data in pbstream and during operation. It reduced the pbstream file size and saved memory consumption during operation.

SirVer:
@ojura I do not agree that this is relevant in this discussion. #508 is about holding scans in memory or not - essentially having an in-memory-asset-writer-light. While @jihoonl is concerned about converting a Cartographer state with the contained map textures into a PGM + yaml.

> I don't get why you got an impression that we are using the different format of serialized data.

It was a misunderstanding of your statement that rosbag is not an option. I understand now that you are not interested in an offline process using canned sensor data at all, instead you specifically only want to use Cartographer state to generate PGM and YAML. I then also understand that you do not perceive the asset writer as an option to you.

I think you would get the best possible quality for your occupancy grid if you were writing a bag with maybe subsampled sensor data and generating your map using the asset writer - it will give equal quality if all submaps are good and higher quality in case any of your submaps in the .pbstream are faulty. It seems to me a more principled approach to your needs that is long term supported and superior in quality.

But to support precisely the use case you add in this new binary, how about we add a service to the occupancy grid node that writes out a PGM and YAML? In your online example you could just keep the OccupancyGrid node running, but not subscribe to its `/map` topic. This means it effectively does not use any CPU. Instead you could call this new rosservice to write out a PGM and YAML occasionally. This would effectively replace `map_saver` for Cartographer and help users of #475 as well. Would that work for you?

jihoonl:
Ok. It looks like I misguided the direction here. I step back and describe the intention behind. 

What I intended with this PR is to introduce a way of introspecting, manipulating and utilising the content of pbstream without bothering cartographer_node(or let's say offline processing without rosbag).  Occupancy grid extraction was the easy first step(and of course to fulfill our usecase) of taking a look the submaps in pbstream. Later, I plan to merge overlapping submaps, drop submaps with no information, and rearrange submap coordination in pbstream.

I understand that this PR has a very specific use case. But, it was to simplify many hassle processes. Having a rosservice in occupancy grid is just a replacement of map_saver which doesn't reduce # of nodes to run to obtain the same output.

Do you concern about opening pbstream data besides `cartographer_node` and `asset_writer`? Or what's your concern? If you don't like the exporter, can you accept the PR that only refactors occupancy grid? Then, I can keep the exporter in our package.

SirVer:
>  I step back and describe the intention behind.

This is much appreciated. Without elaborate context some PRs are really hard to evaluate and review sometimes.

>  Later, I plan to merge overlapping submaps, drop submaps with no information, and rearrange submap coordination in pbstream.

This is interesting and also here I'd like to understand why. It seems superior to me to keep some sensor data around and then reassembling a new state using the best SLAM trajectory in the asset_writer instead of rewiring the online state. Maybe we should discuss this in person or on a VC.

> Do you concern about opening pbstream data besides cartographer_node and asset_writer? Or what's your concern? 

This is one of my concerns. The main ones can be summarized by "no is temporary, yes is forever" ([non-primary-source](https://blog.jessfraz.com/post/the-art-of-closing/)). To make this a bit more fine grained and spell some of them out:

1. Adding a new binary means we have to support it forever. Before we commit to this I want to fully understand that existing solutions can not be simply extended to fulfill the desired use cases.
2. The `.pbstream` data format is non-public, an implementation interface of Cartographer and so far we only support modifying it through the use of Cartographer. We can choose to make it part of Cartographer's public API, but this has similar costs to 1. Again before committing to this I'd like to understand and evaluate the uses cases and see if they can not be expressed on a library level.
3. Related: `.pbstream` is a data format of `cartographer`, not related to ROS. So if a binary derives assets out of it, it should live inside `cartographer`. This would mean moving the drawing code into cartographer and making it part of its public API.
4. Optimizing the cartographer state, i.e. the `.pbstream` seems in scope of Cartographer, but it is not done right now. There I'd like to understand the use cases better.

> If you don't like the exporter, can you accept the PR that only refactors occupancy grid? Then, I can keep the exporter in our package.

Yes, how about for now you pull the drawing code into a library in `cartographer` and make use of it in the occupancy grid node. Then you can have the binary in your own package and we do not bring back the yamlcpp dependency in ROS?

jihoonl:
> This is interesting and also here I'd like to understand why. It seems superior to me to keep some sensor data around and then reassembling a new state using the best SLAM trajectory in the asset_writer instead of rewiring the online state. Maybe we should discuss this in person or on a VC.

We have been using a small size of submap(num_range_data < 30) to reduce errors from odometry. It resulted to have hundreds of submaps with a lot of overlapping area. So, I would like to explore options to manipulate submaps to drop outdated submaps, merge small size submaps, and etc. 

> 1. Adding a new binary means we have to support it forever. Before we commit to this I want to fully understand that existing solutions can not be simply extended to fulfill the desired use cases.

Thanks for the clarification! It helped to understand your concern as a maintainer. I am basically creating PRs that fulfill our use cases and things that can benefit many others as a contributor. Then, letting you to decide how to merge the PR . :smile: :smile: :wink: 

> 2. The .pbstream data format is non-public, an implementation interface of Cartographer and so far we only support modifying it through the use of Cartographer. We can choose to make it part of Cartographer's public API, but this has similar costs to 1. Again before committing to this I'd like to understand and evaluate the uses cases and see if they can not be expressed on a library level.
> 4. Optimizing the cartographer state, i.e. the .pbstream seems in scope of Cartographer, but it is not 
done right now. There I'd like to understand the use cases better.

I think providing an access to `.pbstream` is inevitable for later use such as data introspection, and modification. I can't explain in details since I didn't get to the point to investigate this necessity. But, I have a strong feeling about it.

> 3. Related: .pbstream is a data format of cartographer, not related to ROS. So if a binary derives assets out of it, it should live inside cartographer. This would mean moving the drawing code into cartographer and making it part of its public API.

I was also unsure where it should go. I put it in `cartographer_ros` because pgm and yaml are for ROS though. I agree that the code that touches `.pbstream` would be better to go under `cartographer`.

> Yes, how about for now you pull the drawing code into a library in cartographer and make use of it in the occupancy grid node. Then you can have the binary in your own package and we do not bring back the yamlcpp dependency in ROS?

Ok. sounds good to me. I close this and will open a new PR against cartographer.

ojura:
> Optimizing the cartographer state, i.e. the .pbstream seems in scope of Cartographer, but it is not done right now. There I'd like to understand the use cases better.

This is the exact motivation why I worked on full deserialization of Cartographer's state over at the libcartographer repo.

SirVer:
> Thanks for the clarification! It helped to understand your concern as a maintainer. I am basically creating PRs that fulfill our use cases and things that can benefit many others as a contributor. Then, letting you to decide how to merge the PR . ￼ ￼ ￼

That is fine of course! I hope that you can understand my reluctance too.  My goal is to get everybody what they need. I am sorry if it might occasionally feel like I am actively standing in the way of progress. That is really not my intention.

I think it would help if your goals with any code change are made more clear to us earlier, so that we can advise and iterate earlier on potential approaches.

PR #513, opened 2017-09-27
Follow googlecartographer/cartographer#558

jihoonl:
To prepare initialpose support in cartographer.


jihoonl:
Follow up of googlecartographer/cartographer#558

jihoonl:
I will reopen this if it becomes necessary.

PR #516, opened 2017-09-29
insert resolved topic name in subscribed_topics_

jihoonl:
To be erased while [shutting down subscribers in FinishTrajectory](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/node.cc#L423-L427)

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -155,11 +155,11 @@ void Node::AddExtrapolator(const int trajectory_id,
 void Node::AddSensorSamplers(const int trajectory_id,
                              const TrajectoryOptions& options) {
   CHECK(sensor_samplers_.count(trajectory_id) == 0);
-  sensor_samplers_.emplace(std::piecewise_construct,
-                    std::forward_as_tuple(trajectory_id),
-                    std::forward_as_tuple(options.rangefinder_sampling_ratio,
-                                          options.odometry_sampling_ratio,
-                                          options.imu_sampling_ratio));
+  sensor_samplers_.emplace(
+      std::piecewise_construct, std::forward_as_tuple(trajectory_id),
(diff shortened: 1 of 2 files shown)


damonkohler:
Could you explain why this is necessary? These are only used as internal, unique identifiers.

Mofef:
When a trajectory is stopped and a new one is started again with the same configuration, the respective topics are still marked as taken https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/node.cc#L357

This is because they are inserted to `subscribed_topics_` with the leading slash but are removed from there without (or vice versa). So https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/node.cc#L425 doesn't do anything so far

damonkohler:
I see. However, I think this is the wrong fix. It could break https://github.com/googlecartographer/cartographer_ros/blob/634769e9d02b1a5f07ade23075bdf899ce667d7c/cartographer_ros/cartographer_ros/node.cc#L356 since that depends on ComputeExpectedTopics to return matching results.

Instead, I suggest keeping the internal identifier paired with the actual subscriber here:
https://github.com/googlecartographer/cartographer_ros/blob/634769e9d02b1a5f07ade23075bdf899ce667d7c/cartographer_ros/cartographer_ros/node.h#L159

It's important to be defensive, especially since ros::Subscriber doesn't internally do topic resolution.

jihoonl:
Ah yes. I had a feeling that I missed a place. Ok. I will update the PR as you suggested.

jihoonl:
updated.

SirVer:
I am catching up from my leave. I reviewed this and found the std::pair really confusing. I proposed a change with a `Subscriber` struct and rebased on master. This gave me the opportunity to explain why we keep the 'topic' around in code.

jihoonl:
Yes. looks good to me. It is clearer. thanks.

PR #519, opened 2017-10-04
FetchSubmapTextures returns multiple textures.

gaschler:
This changes the Ros api that multiple textures will be returned,
which are forwarded from the Ros service call.
Adapts usages in RViz to follow this new api.

### DIFF
--- a/cartographer_rviz/cartographer_rviz/drawable_submap.cc
+++ b/cartographer_rviz/cartographer_rviz/drawable_submap.cc
@@ -117,8 +117,8 @@ bool DrawableSubmap::MaybeFetchTexture(ros::ServiceClient* const client) {
   ::cartographer::common::MutexLocker locker(&mutex_);
   // Received metadata version can also be lower if we restarted Cartographer.
   const bool newer_version_available =
-      submap_texture_ == nullptr ||
-      submap_texture_->version != metadata_version_;
+      submap_textures_ == nullptr ||
+      submap_textures_->version != metadata_version_;
   const std::chrono::milliseconds now =
       std::chrono::duration_cast<std::chrono::milliseconds>(
           std::chrono::system_clock::now().time_since_epoch());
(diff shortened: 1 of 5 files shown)


cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -130,15 +130,15 @@ bool DrawableSubmap::MaybeFetchTexture(ros::ServiceClient* const client) {
...
   last_query_timestamp_ = now;
   rpc_request_future_ = std::async(std::launch::async, [this, client]() {
-    std::unique_ptr<::cartographer_ros::SubmapTexture> submap_texture =
-        ::cartographer_ros::FetchSubmapTexture(id_, client);
+    std::unique_ptr<::cartographer_ros::SubmapTextures> submap_textures =
+        ::cartographer_ros::FetchSubmapTextures(id_, client);
     ::cartographer::common::MutexLocker locker(&mutex_);
     query_in_progress_ = false;
-    if (submap_texture != nullptr) {
+    if (submap_textures != nullptr && !submap_textures->textures.empty()) {
```
SirVer: did you not CHECK that empty() can never be? Remove the condition here?

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -168,7 +168,8 @@ void DrawableSubmap::SetAlpha(const double current_tracking_z) {
 
 void DrawableSubmap::UpdateSceneNode() {
   ::cartographer::common::MutexLocker locker(&mutex_);
-  ogre_submap_.Update(*submap_texture_);
+  // TODO(gaschler): Add UI feature to show different textures.
```
SirVer: s/different/all/

cartographer_rviz/cartographer_rviz/drawable_submap.cc
gaschler: Done

cartographer_rviz/cartographer_rviz/drawable_submap.cc
gaschler: Done

PR #522, opened 2017-10-05
Add FrameIdFilteringPointsProcessor

cschuet:
Implements a PointProcessor that filters PointBatches (either by
whitelisting or blacklisting) the frame_id.


SirVer:
There is nothing ROS specific about this PointsProcessor. It belongs in `cartographer` instead of `cartographer_ros`.

PR #524, opened 2017-10-05
No preloading of tf in the assets writer.

### LINKED ISSUE #349: Do not preload all transforms into a tf_buffer in assets_writer_main.
I recently found that tf_buffer was so slow to be the bottleneck in the `offline_node` - especially for longer bags. We worked around this issues in #342 for the `offline_node`, but the same logic is also used in the `assets_writer_main` and should be changed there as well.

wohe:
This follows the logic used in the offline node to only keep
a smaller buffer of tf data. This is necessary for performance.
Fixes #349.

### DIFF
--- a/cartographer_ros/cartographer_ros/assets_writer_main.cc
+++ b/cartographer_ros/cartographer_ros/assets_writer_main.cc
@@ -172,6 +172,10 @@ void Run(const string& pose_graph_filename,
       // otherwise. We make sure that tf_messages are published before any data
       // messages, so that tf lookups always work.
       std::deque<rosbag::MessageInstance> delayed_messages;
+      // We publish tf messages one second earlier than other messages. Under
+      // the assumption of higher frequency tf this should ensure that tf can
+      // always interpolate.
+      const ::ros::Duration kDelay(1.);
       for (const rosbag::MessageInstance& message : view) {
         if (FLAGS_use_bag_transforms && message.isType<tf2_msgs::TFMessage>()) {
           auto tf_message = message.instantiate<tf2_msgs::TFMessage>();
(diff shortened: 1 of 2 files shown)


cartographer_ros/cartographer_ros/assets_writer_main.cc
```diff
@@ -205,25 +168,51 @@ void Run(const string& pose_graph_filename,
...
+            }
+          }
         }
-        if (points_batch != nullptr) {
-          points_batch->trajectory_id = trajectory_id;
-          pipeline.back()->Process(std::move(points_batch));
+
+        while (!delayed_messages.empty() &&
+               delayed_messages.front().getTime() <
+                   message.getTime() - ::ros::Duration(1.)) {
```
cschuet: Is there any significance to this constant? Can you explain why it needs to be 1?

cartographer_ros/cartographer_ros/assets_writer_main.cc
```diff
@@ -205,25 +168,51 @@ void Run(const string& pose_graph_filename,
       const double duration_in_seconds =
           (view.getEndTime() - begin_time).toSec();
 
+      // We make sure that tf_messages are published before any data messages,
+      // so that tf lookups always work and that tf_buffer has a small cache
+      // size - because it gets very inefficient with a large one.
+      std::deque<rosbag::MessageInstance> delayed_messages;
```
cschuet: Please explain the motivation for introducing the delayed messages.

cartographer_ros/cartographer_ros/assets_writer_main.cc
ojura: I was confused about the same thing some time ago. Have a look here:
https://github.com/googlecartographer/cartographer_ros/commit/8b239ded863d4f3fcf4134ce2b575e7f8711df44#commitcomment-22207271

cartographer_ros/cartographer_ros/assets_writer_main.cc
wohe: This is following the offline node. Also see #349.

PR #532, opened 2017-10-10
Visualize both high and low res slices.

gaschler:
Adds checkboxes to the rviz plugin to show high and low
resolution slices.
DrawableSubmap now holds a vector of OgreSlices and
decides which slices are visible.

### DIFF
--- a/cartographer_rviz/cartographer_rviz/drawable_submap.cc
+++ b/cartographer_rviz/cartographer_rviz/drawable_submap.cc
@@ -65,8 +65,7 @@ DrawableSubmap::DrawableSubmap(const ::cartographer::mapping::SubmapId& id,
                           .arg(id.trajectory_id)
                           .arg(id.submap_index)
                           .toStdString()),
-      last_query_timestamp_(0),
-      slices_visibility_(kNumberOfSlicesPerSubmap, true) {
+      last_query_timestamp_(0) {
   for (int slice_index = 0; slice_index < kNumberOfSlicesPerSubmap;
        ++slice_index) {
     ogre_slices_.emplace_back(::cartographer::common::make_unique<OgreSlice>(
@@ -174,9 +173,8 @@ void DrawableSubmap::SetAlpha(const double current_tracking_z) {
 }
 
 void DrawableSubmap::SetSliceVisibility(size_t slice_index, bool visible) {
-  if (slice_index < slices_visibility_.size()) {
-    slices_visibility_[slice_index] = visible;
-  }
+  CHECK(slice_index < ogre_slices_.size());
+  ogre_slices_[slice_index]->SetVisibility(visible);
   ToggleVisibility();
 }
 
@@ -191,11 +189,8 @@ void DrawableSubmap::UpdateSceneNode() {
 }
 
 void DrawableSubmap::ToggleVisibility() {
-  for (size_t slice_index = 0; slice_index < ogre_slices_.size() &&
-                               slice_index < slices_visibility_.size();
-       ++slice_index) {
-    ogre_slices_[slice_index]->SetVisibility(visibility_->getBool() &&
-                                             slices_visibility_[slice_index]);
+  for (auto& ogre_slice : ogre_slices_) {
+    ogre_slice->UpdateOgreNodeVisibility(visibility_->getBool());
   }
   display_context_->queueRender();
 }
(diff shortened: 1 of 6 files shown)


cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -162,19 +167,36 @@ void DrawableSubmap::SetAlpha(const double current_tracking_z) {
...
     current_alpha_ = target_alpha;
   }
-  ogre_slice_.SetAlpha(current_alpha_);
+  for (auto& slice : ogre_slices_) {
+    slice->SetAlpha(current_alpha_);
+  }
   display_context_->queueRender();
 }
 
+void DrawableSubmap::SetSliceVisibility(size_t slice_index, bool visible) {
```
SirVer: instead of if use CHECK? it seems a programming error if 'slice_index' is out of range.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -162,19 +167,36 @@ void DrawableSubmap::SetAlpha(const double current_tracking_z) {
...
+    slices_visibility_[slice_index] = visible;
+  }
+  ToggleVisibility();
+}
+
 void DrawableSubmap::UpdateSceneNode() {
   ::cartographer::common::MutexLocker locker(&mutex_);
-  // TODO(gaschler): Add UI feature to show all textures.
-  ogre_slice_.Update(submap_textures_->textures[0]);
+  for (size_t slice_index = 0; slice_index < ogre_slices_.size() &&
```
SirVer: do we ever expect to get a submap_textures_.size() != ogre_slices_? I think this would be invalid data, dropped earlier. So CHECK this here?

cartographer_rviz/cartographer_rviz/drawable_submap.h
```diff
@@ -105,6 +109,7 @@ class DrawableSubmap : public QObject {
       GUARDED_BY(mutex_);
   float current_alpha_ = 0.f;
   std::unique_ptr<::rviz::BoolProperty> visibility_;
+  std::vector<bool> slices_visibility_;
```
SirVer: It is an invariant that slices_visibility_.size() == ogre_slices_.size(), right? Encode this, by pulling out a private struct `Slice` or move it into OgreSlice.

cartographer_rviz/cartographer_rviz/ogre_slice.cc
```diff
@@ -142,4 +144,8 @@ void OgreSlice::SetAlpha(const float alpha) {
   parameters->setNamedConstant("u_alpha", alpha);
 }
 
+void OgreSlice::SetVisibility(bool visible) {
```
SirVer: Visibility is a strongly overloaded term here. You already track the 'visibility' bit for each slice in DrawableSubmap. Maybe SetOgreNodeVisibility. Alternatively, you could move the 'visibility' bit into this class from DrawableSubmap and have a `UpdateNodeVisibility(bool submap_visibility)`.

cartographer_rviz/cartographer_rviz/ogre_slice.h
```diff
@@ -55,8 +55,11 @@ class OgreSlice {
   // Changes the opacity of the submap to 'alpha'.
   void SetAlpha(float alpha);
 
+  void SetVisibility(bool visible);
+
  private:
   const ::cartographer::mapping::SubmapId id_;
+  int slice_id_;
```
SirVer: const int, since it never changes?

cartographer_rviz/cartographer_rviz/drawable_submap.cc
wohe: 2D will continue to send one slice.

cartographer_rviz/cartographer_rviz/drawable_submap.cc
```diff
@@ -174,9 +173,8 @@ void DrawableSubmap::SetAlpha(const double current_tracking_z) {
 }
 
 void DrawableSubmap::SetSliceVisibility(size_t slice_index, bool visible) {
-  if (slice_index < slices_visibility_.size()) {
-    slices_visibility_[slice_index] = visible;
-  }
+  CHECK(slice_index < ogre_slices_.size());
```
SirVer: shorter: ogre_slices_.at(slice_index)->Set...

gaschler: Done

Issue #529, opened 2017-10-09
Build a rosbag_validate


SirVer:
Many of the tuning issues we receive in Cartographer are easy to explain by looking through the bags provided. Below I list some issues that could have benefited from a binary that runs through bags and does automatic checks for validity and likeliness of data as wells as obvious configuration errors.

Some examples for inspiration what to check for with example issues where this showed:

- [x] timing (#511, #317, googlecartographer/cartographer#377, googlecartographer/cartographer#193), most of these can be tracked with a histogram + min/max of delta timestamps for each sensor.
- [x] significant difference in a) `rosbag::MessageInstance::getTime` and b) `msg->header.stamp`. The problem is that Cartographer will generate pose timestamps from b) since it is purely data driven, but rViz will want to visualize poses for `ros::Time::Now()` which is published by rosbag and `cartographer_offline_node` using a). (#488, https://github.com/googlecartographer/cartographer/issues/447)
- [x] incomplete IMU data (#424). Especially wrong units (G instead of m/s^2 or gravity being removed)
- [x] fastest sensor is still very slow (#372)
- [x] frames fighting: TF published from Cartographer are also in the bag (#356).
- [x] scans are never changing shape (#356)
- [ ] ROS data type validation: For example invalid laser messages (#354, #270 ), invalid IMU messages (#273, #268, #462), [odometry data with invalid quaternions](https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/google-cartographer/Jtml5Kf9Cg4/hM63bqFfBgAJ).
 - [ ] header time stamps not strictly increasing (https://github.com/googlecartographer/cartographer_ros/issues/699)

ojura:
IIRC, Cartographer has a zero velocity assumption at the beginning of the bag, so checking for that might also be useful.

MichaelGrupp:
Should be done now in `cartographer_rosbag_validate`.

PR #536, opened 2017-10-13
Adds a rosbag_validate binary.

SirVer:
So far it creates statistics over timing information of the sensor data that Cartographer cares about. It also reports if time jumps backwards for a sensor.
Also update the GitHub issue template to ask user to run this binary when reporting issues.

I verified that this tool is sufficient to analyze all the timing related issues mentioned in #529.


SirVer:
@drigz Could you review for English?

ojura:
You have a typo in the PR title :-)

drigz:
Approval for English.

SirVer:
@gaschler Andre, I'll await your code review before merging.

cartographer_ros/cartographer_ros/rosbag_wtf_main.cc
```diff
@@ -0,0 +1,121 @@
...
+#include "sensor_msgs/Imu.h"
+#include "sensor_msgs/LaserScan.h"
+#include "sensor_msgs/MultiEchoLaserScan.h"
+#include "sensor_msgs/PointCloud2.h"
+#include "tf2_eigen/tf2_eigen.h"
+#include "tf2_msgs/TFMessage.h"
+#include "tf2_ros/buffer.h"
+#include "urdf/model.h"
+
+DEFINE_string(bag_filename, "", "Bag to process");
```
gaschler: "Bag to process." seems more standard compared to the other binaries.

cartographer_ros/cartographer_ros/CMakeLists.txt
```diff
@@ -66,3 +66,14 @@ install(TARGETS cartographer_occupancy_grid_node
   LIBRARY DESTINATION ${CATKIN_PACKAGE_LIB_DESTINATION}
   RUNTIME DESTINATION ${CATKIN_PACKAGE_BIN_DESTINATION}
 )
+
+google_binary(cartographer_rosbag_wtf
+  SRCS
+    rosbag_wtf_main.cc
+)
+
+install(TARGETS cartographer_rosbag_wtf
```
gaschler: Maybe a different name (for binary and source) sounds friendlier, but I leave the decision to you.

cartographer_ros/cartographer_ros/rosbag_wtf_main.cc
```diff
@@ -0,0 +1,121 @@
...
+                      "acquired from the sensor.";
+      }
+      entry.histogram.Add(delta_t_sec);
+      entry.last_timestamp = time;
+    }
+  }
+  bag.close();
+
+  constexpr int kNumBucketsForHistogram = 10;
+  for (const auto& entry_pair : timestamp_data) {
```
gaschler: Looks like this will output every timestamp even if everything is fine, but I guess this is intended for this first version. 
(My understanding was the output should be short enough to paste it into an issue.)

SirVer: I do not fully understand what you are asking. It outputs an ASCII histogram of each sensors timestamp deltas. This is still short enough to be pasted into an issue, but you are right that it might get longer over time.

I changed the issue template to ask for a gist of the output instead.

cartographer_ros/cartographer_ros/CMakeLists.txt
SirVer: I chose the name for symmetry with `roswtf`, but I am not a fan of the name myself. After offline discussion the benefit of the symmetry of naming is less important than the unpleasant connotations of the vulgar name. I therefore renamed the tool.

PR #556, opened 2017-10-25
Use AccumulatedRangeDataCallback for trajectory states

ojura:
Follows googlecartographer/cartographer#574

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -168,25 +168,37 @@ void Node::PublishTrajectoryStates(const ::ros::WallTimerEvent& timer_event) {
     const auto& trajectory_state = entry.second;
 
     auto& extrapolator = extrapolators_.at(entry.first);
+    const auto& last_range_data = trajectory_state.data.last_range_data;
+    const auto& last_constant_data = trajectory_state.data.last_constant_data;
+    CHECK(last_constant_data != nullptr);
+    CHECK(last_range_data != nullptr);
+
+    const auto& time = last_constant_data->time;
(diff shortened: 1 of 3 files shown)


cartographer_ros/cartographer_ros/node.cc
```diff
@@ -168,25 +168,37 @@ void Node::PublishTrajectoryStates(const ::ros::WallTimerEvent& timer_event) {
...
     // We only publish a point cloud if it has changed. It is not needed at high
     // frequency, and republishing it would be computationally wasteful.
-    if (trajectory_state.pose_estimate.time != extrapolator.GetLastPoseTime()) {
+    if (time != extrapolator.GetLastPoseTime()) {
+      const carto::sensor::PointCloud original_point_cloud =
+          carto::sensor::TransformPointCloud(
+              last_range_data->returns,
+              (last_constant_data->local_pose
+               // THIS IS CURRENTLY BROKEN FOR 3D. THE FOLLOWING MULTIPLICATION
+               // SHOULD BE APPLIED ONLY FOR 2D!
```
ojura: As noted in the code comment. The range data in 2D and 3D does not currently have the same relationship with `last_constant_data->local_pose`.

wohe: This is a bad idea. The only reason for the existence of the gravity aligned frame is that it is necessary to express certain rigid transforms using `Rigid2`. Other than that, it should not be used.

ojura: I am aware it is bad idea, we need to discuss what to change in libcartographer to resolve this.

PR #558, opened 2017-10-26
Don't `set -o verbose` in ros_entrypoint.sh

drigz:
All it does is source some large scripts, which results in lots of
logspam when the container starts up. It then runs exec, after which the
`verbose` setting has no effect.


SirVer:
lgtm. Need to rebase before merging.

PR #566, opened 2017-10-30
Changes in BQ client API and constraints stats.

cschuet:


jenkins/worker.py
```diff
@@ -31,6 +30,36 @@ def extract(self, inp):
       return {}
     return match.groupdict()
 
+  def extract_last_occurence(self, inp):
```
damonkohler: inp? Please find a better name.

PR #575, opened 2017-11-06
Move Rosbag IO into background thread in offline_node_main.

SirVer:
Timings using `/usr/bin/time roslaunch cartographer_ros offline_backpack_3d.launch bag_filenames:=$HOME/Downloads/b3-2015-12-10-13-55-20.bag` after removing `occupancy_grid_node`, rViz, and setting `required="True"` on the node.

master:
Elapsed wall clock time: 232.837 s (100%)
Elapsed CPU time: 386.774 s (100%)
Peak memory usage: 1515016 KiB (100%)
372.32user (100%) 68.22system (100%) 4:04.33 (100%) elapsed 180%CPU k

this branch:
Elapsed wall clock time: 211.506 s (90.8%)
Elapsed CPU time: 386.884 s (100%)
Peak memory usage: 1554836 KiB (102.6%)
375.02user (100.8%) 62.09system (100%) 3:43.10elapsed (94.39%) 195%CPU

### DIFF
--- a/cartographer_ros/cartographer_ros/offline_node_main.cc
+++ b/cartographer_ros/cartographer_ros/offline_node_main.cc
@@ -24,6 +24,7 @@
 #include <string>
 #include <vector>
 
+#include "cartographer/common/blocking_queue.h"
 #include "cartographer/common/make_unique.h"
 #include "cartographer/common/port.h"
 #include "cartographer_ros/node.h"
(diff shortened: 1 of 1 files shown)


SirVer:
@cschuet @wohe Is getting 10% of your lifetime back worth the increase in code and complexity? I am quite torn - I was also hoping for higher gains.

SirVer:
@wohe I tried the lambda approach in https://github.com/SirVer/cartographer_ros/tree/01_multithread_use_lambdas, unfortunately it takes ~150% of current master, so is much slower.

SirVer:
After some more experiments I am not convinced that this will have the impact that I was hoping for. I think the rfc 0, i.e. moving towards a gRPC based solution and wrapping this in ROS will have similar results (i.e. prereading in the ROS thread will happen) and should be cleaner. Abandoning this for the moment.

Issue #577, opened 2017-11-06
Add a launch file and/or documentation for writing a ROS map


ojura:
This should be documented and/or turned into an example:
https://github.com/googlecartographer/cartographer_ros/issues/553#issuecomment-338669056

MichaelGrupp:
`cartographer_pbstream_to_ros_map` is now the simplest way to do it and seems to be used by people. Documentation improvements are always welcome though :)

PR #580, opened 2017-11-07
Add 3D localization demo.

wohe:
Adds configuration files and launch files for 3D backpack pure
localization.

Moves most demos to a separate documentation page.

### DIFF
--- a/docs/source/demo.rst
+++ b/docs/source/demo.rst
@@ -1,0 +1,79 @@
+.. Copyright 2016 The Cartographer Authors
+
+.. Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+..      http://www.apache.org/licenses/LICENSE-2.0
+
+.. Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+
+=====
+Demos
+=====
+
+Pure localization
+=================
+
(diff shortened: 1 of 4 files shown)


docs/source/demo.rst
```diff
@@ -0,0 +1,79 @@
...
+    roslaunch cartographer_ros demo_backpack_2d_localization.launch \
+       map_filename:=${HOME}/Downloads/b2-2016-04-05-14-44-52.bag.pbstream \
+       bag_filename:=${HOME}/Downloads/b2-2016-04-27-12-31-41.bag
+
+    # Pure localization demo in 3D: We use 2 different 3D bags from the Deutsche
+    # Museum. The first one is used to generate the map, the second to run
+    # pure localization.
+    wget -P ~/Downloads https://storage.googleapis.com/cartographer-public-data/bags/backpack_3d/b3-2016-04-05-13-54-42.bag
+    wget -P ~/Downloads https://storage.googleapis.com/cartographer-public-data/bags/backpack_3d/b3-2016-04-05-15-52-20.bag
+    # Generate the map: Run the next command, wait until cartographer_offline_node finishes.
```
ojura: Assets are not written at the end of offline node anymore, right? Now you get a .pbstream and need to run assets writer. See #577.

wohe: Correct. This does not write assets, but it writes the `.pbstream` which is all we need. You do not have to run the assets writer for this demo.

ojura: Okay, but that means that the "Generate the map" part of the command is not accurate.

gaschler: We localize in a map, and that map is in a pbstream file (not an image or other asset). Perhaps the term "map" is overloaded here but I think this is fine.

ojura: Oh right, sorry, I missed the context.

PR #584, opened 2017-11-08
Apply topic remappings to start_trajectory_node.

ojura:
This enables us to easily tell Cartographer node which topic names to use by remapping topics of `start_trajectory_node`.

AFAIK previously the only way to do this was by crafting a custom `StartTrajectory` service request with the desired topic names, right? How else could you tell Cartographer to use `scan0` for the first trajectory and `scan1` for the second?


cartographer_ros/cartographer_ros/start_trajectory_main.cc
```diff
@@ -59,11 +59,14 @@ bool Run() {
           kStartTrajectoryServiceName);
   cartographer_ros_msgs::StartTrajectory srv;
   srv.request.options = ToRosMessage(LoadOptions());
-  srv.request.topics.laser_scan_topic = kLaserScanTopic;
-  srv.request.topics.multi_echo_laser_scan_topic = kMultiEchoLaserScanTopic;
-  srv.request.topics.point_cloud2_topic = kPointCloud2Topic;
-  srv.request.topics.imu_topic = kImuTopic;
-  srv.request.topics.odometry_topic = kOdometryTopic;
+  srv.request.topics.laser_scan_topic =
```
SirVer: Please add `true /* explain boolean meaning */.

ojura: Just for the first one, right?

SirVer:
lgtm. Thanks!

ojura:
@SirVer I just did some more digging around out of curiosity and it kind of seems like `node_handle.resolveName(topic)` might be a safer (more general) choice instead of `ros::names::resolve(topic)`, as used in this PR. `NodeHandle::resolveName()` is already used in the offline node for the same purpose [here](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros/cartographer_ros/offline_node_main.cc#L103). Also, I missed that the remapping bool can be omitted from the call, since it defaults to `true`. But perhaps it's good that it's there, to make the intent clearer.

Before submitting the PR, I have tried running e.g. `rosrun cartographer_ros cartographer_start_trajectory .... scan:=some_other_scan` and of course it works as expected with the code in this PR, even when running the helper node within a namespace. But, looking into ROS source [here](http://docs.ros.org/kinetic/api/roscpp/html/node__handle_8cpp_source.html#l00215), I am not completely sure why it works. The first thought that came to me was that the remapping given to the node on command line/in a launch file is a _local_ remapping;  but since calling `ros::names::resolve()` works, it seems like this intuition is wrong and it's a global remapping, when looking within the context of `NodeHandle` CPP source.

What is then a local remapping? it might be a node-handle specific remapping. It seems that you can explicitly pass these additional _local_ remappings [when constructing the node handle](http://docs.ros.org/kinetic/api/roscpp/html/node__handle_8cpp_source.html#l00073), and it looks like whether these local remappings are applied is the biggest difference between `NodeHandle::resolveName()` and `ros::names::resolve()`. However, since we're not passing any remappings during `NodeHandle` construction (like anyone sane), perhaps we can ignore this difference and keep this as it is.

To summarize: this might be nitpicking a bit, but I was just wondering what's the difference between the two resolution functions. It looks like we can leave everything as it is, if I got everything right in the paragraph above, but perhaps you might want to double-check.

SirVer:
@ojura I am also not sure. We use node_handle.resolve in other places, so maybe for consistency we should change it to that.

PR #588, opened 2017-11-10
Remove io component of PCL.

ojura:
The `io` component of PCL brings in VTK, which in turn creates a lot of CMake pollution (such as bringing in Qt, which is completely unnecessary for `cartographer_ros`).


SirVer:
Thanks!

@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user SirVer. Merge queue now has a length of 1.

PR #610, opened 2017-11-23
Add initial_pose in start_trajectory_main.cc Fixes #579

### LINKED ISSUE #579: Initial Pose Wiring
The initial pose implementation in https://github.com/googlecartographer/cartographer/pull/606 is ready to merge but as of yet untested.

Initial poses need to be wired up in cartographer_ros (probably have an initial_pose topic) and tested that they work correctly.

jihoonl:
Fixes #579 
Related to googlecartographer/cartographer#606

@damienrg @cschuet I followed most of the comments in googlecartographer/cartographer#642 except timestamp. Receiving timestamp sounds weird to me because trajectory should not start in past timestamp or future timestamp.

### DIFF
--- a/cartographer_ros/cartographer_ros/start_trajectory_main.cc
+++ b/cartographer_ros/cartographer_ros/start_trajectory_main.cc
@@ -51,14 +51,13 @@ TrajectoryOptions LoadOptions() {
   auto lua_parameter_dictionary =
       cartographer::common::LuaParameterDictionary::NonReferenceCounted(
           code, std::move(file_resolver));
-  if(!FLAGS_initial_pose.empty()) {
+  if (!FLAGS_initial_pose.empty()) {
     auto initial_trajectory_pose =
         cartographer::common::LuaParameterDictionary::NonReferenceCounted(
             "return " + FLAGS_initial_pose, std::move(file_resolver));
     return CreateTrajectoryOptions(lua_parameter_dictionary.get(),
                                    initial_trajectory_pose.get());
-  }
-  else {
+  } else {
     return CreateTrajectoryOptions(lua_parameter_dictionary.get());
   }
 }
--- a/cartographer_ros/cartographer_ros/start_trajectory_main.cc
+++ b/cartographer_ros/cartographer_ros/start_trajectory_main.cc
@@ -52,9 +52,14 @@ TrajectoryOptions LoadOptions() {
       cartographer::common::LuaParameterDictionary::NonReferenceCounted(
           code, std::move(file_resolver));
   if (!FLAGS_initial_pose.empty()) {
+    auto initial_trajectory_pose_file_resolver =
+        cartographer::common::make_unique<
+            cartographer::common::ConfigurationFileResolver>(
+            std::vector<std::string>{FLAGS_configuration_directory});
     auto initial_trajectory_pose =
         cartographer::common::LuaParameterDictionary::NonReferenceCounted(
-            "return " + FLAGS_initial_pose, std::move(file_resolver));
+            "return " + FLAGS_initial_pose,
+            std::move(initial_trajectory_pose_file_resolver));
     return CreateTrajectoryOptions(lua_parameter_dictionary.get(),
                                    initial_trajectory_pose.get());
   } else {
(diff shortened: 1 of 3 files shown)


cartographer_ros/cartographer_ros/trajectory_options.cc
```diff
@@ -69,6 +72,29 @@ TrajectoryOptions CreateTrajectoryOptions(
...
+CreateInitialTrajectoryPose(
+    ::cartographer::common::LuaParameterDictionary* lua_parameter_dictionary) {
+  ::cartographer::mapping::proto::InitialTrajectoryPose pose;
+  pose.set_to_trajectory_id(
+      lua_parameter_dictionary->GetNonNegativeInt("to_trajectory_id"));
+  *pose.mutable_relative_pose() =
+      cartographer::transform::ToProto(cartographer::transform::FromDictionary(
+          lua_parameter_dictionary->GetDictionary("relative_pose").get()));
+  pose.set_timestamp(
+      cartographer::common::ToUniversal(FromRos(ros::Time::now())));
```
cschuet: You want to read the timestamp from the command line. The timestamp does not specify when the new trajectory started, rather it indicates at what time you you want to take the global pose from the to_trajectory apply your relative pose and make that your starting position.

damienrg: Yes but when you want to use it is so that not simple. I think that starting a new trajectory with initial pose does not play well with start_trajectory_main because you must have access to the pose of the trajectory at a certain time. Maybe I am wrong but for me start_trajectory is used "interactively" so it means that:
- you have cartographer_ros running
- you see the construction of the map and at particular time you decide to start a new trajectory
In this case providing a timestamp is cumbersome, you have to find the current time and use it to start the new trajectory with an identity matrix.

To support what @cschuet said and the use case above I think that the timestamp should be optional and if not provided set to current ros time.

cschuet: Ok, I see that use case makes sense to me. Also I am not super familiar with ROS. I think my concern was that taking the current time may lead to unexpected behavior when the relation between recorded timestamps and system time is not what you expect. But reading up on `ros::Time::now()` it seems that within a rosbag playback it returns the current time *in the simulation*. If that's the case I am ok with having it default to `ros::Time::now()` when no timestamp is provided.

jihoonl: Yes, `ros::Time::now()` was to compatible with ros simulated time from rosbag.

Thank you guys for review. I will update the PR based on your comment.

jihoonl:
Ok. updated. ready for review.

damienrg:
Looks good to me.
Maybe edit the first comment as it will be in the git history?

jihoonl:
Ok. I have added in the title. It will be the actual commit message I think.

cartographer_ros/cartographer_ros/start_trajectory_main.cc
```diff
@@ -49,7 +51,15 @@ TrajectoryOptions LoadOptions() {
   auto lua_parameter_dictionary =
       cartographer::common::LuaParameterDictionary::NonReferenceCounted(
           code, std::move(file_resolver));
-  return CreateTrajectoryOptions(lua_parameter_dictionary.get());
+  if (!FLAGS_initial_pose.empty()) {
+    auto initial_trajectory_pose =
+        cartographer::common::LuaParameterDictionary::NonReferenceCounted(
+            "return " + FLAGS_initial_pose, std::move(file_resolver));
```
cschuet: Are you not moving std::move(file_resolver) a second time (see l. 53)?

jihoonl: ah yes you are right. I wasn't sure how to handle here since LuaParameterDictionary isn't constructable without file resolver. Should I pass nullptr?

cschuet: I know you are not using the `file_resolver` and `nullptr` would not crash but please just construct another `file_resolver` inside the if-block and std::move it into your LuaParameterDictionary.

Please run this code once with and w/o flag to make sure it works as expeceted.

jihoonl: I have confirmed it is working for both cases. And added file resolver for initial pose lua dict.

cschuet:
@jihoonl Thanks for the PR! Ship it!

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #616, opened 2017-12-04
Follow googlecartographer/cartographer/pull/724.

gaschler:


cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -97,7 +78,11 @@ int MapBuilderBridge::AddTrajectory(
     const std::unordered_set<std::string>& expected_sensor_ids,
     const TrajectoryOptions& trajectory_options) {
   const int trajectory_id = map_builder_.AddTrajectoryBuilder(
-      expected_sensor_ids, trajectory_options.trajectory_builder_options);
+      expected_sensor_ids, trajectory_options.trajectory_builder_options,
+      ::std::bind(&MapBuilderBridge::OnLocalSlamResult, this,
+                  ::std::placeholders::_1, ::std::placeholders::_2,
+                  ::std::placeholders::_3, ::std::placeholders::_4,
+                  ::std::placeholders::_5));
```
cschuet: For posteritie's sake:
1. We discussed the possibility to directly pass an anonymous function here (like originally in the const'or), but disliked this approach since the logic is complex enough that it deserves to live in a function.
2. We discussed adding something template magic to make perfect-forwarding play nice with member functions, but it seemed like an overkill for just one case.
3. We discussed creating an anonymous function which invokes `OnLocalSlamResult` but this would mean enumerating all arguments twice.

PR #632, opened 2017-12-20
 Implement cartographer_grpc_node.

cschuet:


cartographer_ros/cartographer_ros/cartographer_grpc/node_grpc_main.cc
```diff
@@ -26,7 +28,9 @@ DEFINE_string(configuration_directory, "",
 DEFINE_string(configuration_basename, "",
               "Basename, i.e. not containing any directory prefix, of the "
               "configuration file.");
-DEFINE_string(map_filename, "", "If non-empty, filename of a map to load.");
+DEFINE_string(server_address, "localhost:50051",
```
gaschler: Is there a way to get this default value from the original? Defining this string here is likely to get out of sync.

cschuet: There no easy was as discussed: on the server side the default is set in the lua file the server options are parsed from. cartographer_node_grpc however is not configured using lua files which is why there is no other place in the C++ code where this constant appears.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #634, opened 2017-12-26
Constraints visualization: Separate inter constraints between separate trajectories

ojura:
When troubleshooting global localization between different trajectories, it can be useful to single out inter-trajectory constraints.

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -247,15 +247,33 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintList() {
   // visible.
   residual_intra_marker.pose.position.z = 0.1;
 
-  visualization_msgs::Marker constraint_inter_marker = constraint_intra_marker;
-  constraint_inter_marker.id = marker_id++;
-  constraint_inter_marker.ns = "Inter constraints";
-  constraint_inter_marker.pose.position.z = 0.1;
+  visualization_msgs::Marker constraint_inter_same_trajectory_marker =
+      constraint_intra_marker;
+  constraint_inter_same_trajectory_marker.id = marker_id++;
+  constraint_inter_same_trajectory_marker.ns =
+      "Inter constraints, same trajectories";
+  constraint_inter_same_trajectory_marker.pose.position.z = 0.1;
 
-  visualization_msgs::Marker residual_inter_marker = constraint_intra_marker;
-  residual_inter_marker.id = marker_id++;
-  residual_inter_marker.ns = "Inter residuals";
-  residual_inter_marker.pose.position.z = 0.1;
+  visualization_msgs::Marker residual_inter_same_trajectory_marker =
+      constraint_intra_marker;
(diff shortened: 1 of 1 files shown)


cschuet:
@ojura Do you want to get this merged? If so can you provide a screenshot of what it does? I am having trouble understanding what you are achieving with this.

ojura:
Yeah, I do. I will provide you with a screenshot later (and update the PR), but in a nutshell: if you're debugging global localization of one trajectory to another (i.e. solving the kidnapped robot problem), you hope that the global constraint builder will find valid constraints between the two trajectories and thus properly align them together (instead of having the second rajectory float around the first trajectory, unconnected, or having them aligned improperly due to a false global localization match). I was having problems like this, so while tuning, I wanted to see the constraints generated between the two trajectories, and _not_ the regular inter loop closure constraints generated within one trajectory with respect to itself. The latter can be numerous when compared to the former, and they are both rendered in shiny yellow colours. By separating them in different namespaces, you can choose to display only the constraints between two different trajectories.

gaschler:
@ojura, this is a little bit related to https://github.com/googlecartographer/cartographer/issues/656, would you like to also take over the visualization how far global SLAM has come?

ojura:
I updated the PR.

#656 sounds useful. I'd rather not claim it and then not do it due to lack of time, but I'll keep it in mind.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -247,15 +247,33 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintList() {
...
-
-  visualization_msgs::Marker residual_inter_marker = constraint_intra_marker;
-  residual_inter_marker.id = marker_id++;
-  residual_inter_marker.ns = "Inter residuals";
-  residual_inter_marker.pose.position.z = 0.1;
+  visualization_msgs::Marker constraint_inter_same_trajectory_marker =
+      constraint_intra_marker;
+  constraint_inter_same_trajectory_marker.id = marker_id++;
+  constraint_inter_same_trajectory_marker.ns =
+      "Inter constraints, same trajectories";
```
gaschler: same trajector*y*, also in line 261

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -278,8 +296,15 @@ visualization_msgs::MarkerArray MapBuilderBridge::GetConstraintList() {
       color_residual.a = 1.0;
       color_residual.r = 1.0;
     } else {
-      constraint_marker = &constraint_inter_marker;
-      residual_marker = &residual_inter_marker;
+      if (constraint.node_id.trajectory_id ==
+          constraint.submap_id.trajectory_id) {
+        constraint_marker = &constraint_inter_same_trajectory_marker;
```
gaschler: How about making these constraints bright yellow and the cross-trajectory ones bright orange? What do you think?

cschuet:
@wally-the-cartographer merge

cschuet:
@wally-the-cartographer merge

PR #636, opened 2017-12-26
Simultaneous offline multi trajectories

ojura:
RFC=[0009](https://github.com/googlecartographer/rfcs/pull/4)

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -122,7 +122,7 @@ Node::Node(
       &Node::PublishConstraintList, this));
 }
 
-Node::~Node() {}
+Node::~Node() { FinishAllTrajectories(); }
 
 ::ros::NodeHandle* Node::node_handle() { return &node_handle_; }
 
--- a/cartographer_ros/cartographer_ros/offline_node.cc
+++ b/cartographer_ros/cartographer_ros/offline_node.cc
@@ -167,7 +167,7 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
        ++current_bag_index) {
     const std::string& bag_filename = bag_filenames.at(current_bag_index);
     if (!::ros::ok()) {
-      break;
+      return;
     }
     std::unordered_set<std::string> current_expected_sensor_topics;
     if (specified_sensor_topics.at(current_bag_index).empty()) {
@@ -206,7 +206,7 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
     const bool bag_has_more_messages = std::get<2>(next_msg_tuple);
 
     if (!::ros::ok()) {
-      break;
+      return;
     }
     const std::string topic =
         node.node_handle()->resolveName(msg.getTopic(), false /* resolve */);
--- a/cartographer_ros/cartographer_ros/offline_node.cc
+++ b/cartographer_ros/cartographer_ros/offline_node.cc
@@ -168,6 +168,7 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
 
   std::unordered_map<int, std::unordered_set<std::string>>
       expected_sensor_topics;
+  std::unordered_map<int, int> bag_id_to_trajectory_id;
   PlayableBagMultiplexer playable_bag_multiplexer;
   for (size_t current_bag_index = 0; current_bag_index < bag_filenames.size();
        ++current_bag_index) {
@@ -191,17 +192,14 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
                   .second);
       }
     }
-    const int trajectory_id =
-        node.AddOfflineTrajectory(current_expected_sensor_topics,
-                                  trajectory_options.at(current_bag_index));
     CHECK(expected_sensor_topics
               .emplace(std::piecewise_construct,
-                       std::forward_as_tuple(trajectory_id),
+                       std::forward_as_tuple(current_bag_index),
                        std::forward_as_tuple(current_expected_sensor_topics))
               .second);
 
     playable_bag_multiplexer.AddPlayableBag(PlayableBag(
-        bag_filename, trajectory_id, ros::TIME_MIN, ros::TIME_MAX, kDelay,
+        bag_filename, current_bag_index, ros::TIME_MIN, ros::TIME_MAX, kDelay,
         [&](const rosbag::MessageInstance& msg) {
           if (msg.isType<tf2_msgs::TFMessage>()) {
             if (FLAGS_use_bag_transforms) {
@@ -231,15 +229,30 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
   while (playable_bag_multiplexer.IsMessageAvailable()) {
     const auto next_msg_tuple = playable_bag_multiplexer.GetNextMessage();
     const rosbag::MessageInstance& msg = std::get<0>(next_msg_tuple);
-    const int trajectory_id = std::get<1>(next_msg_tuple);
+    const int bag_index = std::get<1>(next_msg_tuple);
     const bool bag_has_more_messages = std::get<2>(next_msg_tuple);
+
+    int trajectory_id;
+    // Lazily add trajectories only when the first message arrives.
+    if (bag_id_to_trajectory_id.count(bag_index) == 0) {
+      trajectory_id =
+          node.AddOfflineTrajectory(expected_sensor_topics.at(bag_index),
+                                    trajectory_options.at(bag_index));
+      CHECK(bag_id_to_trajectory_id
+                .emplace(std::piecewise_construct,
+                         std::forward_as_tuple(bag_index),
+                         std::forward_as_tuple(trajectory_id))
+                .second);
+    } else {
+      trajectory_id = bag_id_to_trajectory_id.at(bag_index);
+    }
 
     if (!::ros::ok()) {
       return;
     }
     const std::string topic =
         node.node_handle()->resolveName(msg.getTopic(), false /* resolve */);
-    if (expected_sensor_topics.at(trajectory_id).count(topic) == 0) {
+    if (expected_sensor_topics.at(bag_index).count(topic) == 0) {
       continue;
     }
 
--- a/cartographer_ros/cartographer_ros/offline_node.cc
+++ b/cartographer_ros/cartographer_ros/offline_node.cc
@@ -35,15 +35,19 @@ DEFINE_string(configuration_directory, "",
               "First directory in which configuration files are searched, "
               "second is always the Cartographer installation to allow "
               "including files from there.");
-DEFINE_string(configuration_basenames, "",
-              "Basenames, i.e. not containing any directory prefix, of the "
-              "configuration files for each trajectory.");
+DEFINE_string(
+    configuration_basenames, "",
+    "Comma-separated list of basenames, i.e. not containing any "
+    "directory prefix, of the configuration files for each trajectory. "
+    "The first configuration file will be used for node options."
+    "If less configuration files are specified than trajectories, the "
+    "first file will be used the remaining trajectories.");
 DEFINE_string(
     bag_filenames, "",
     "Comma-separated list of bags to process. One bag per trajectory.");
 DEFINE_string(urdf_filenames, "",
-              "One or more URDF files that contains static links for your "
-              "sensor configuration.");
+              "Comma-separated list of one or more URDF files that contain "
+              "static links for the sensor configuration(s).");
 DEFINE_bool(use_bag_transforms, true,
             "Whether to read, use and republish transforms from bags.");
 DEFINE_string(pbstream_filename, "",
@@ -52,10 +56,14 @@ DEFINE_bool(keep_running, false,
             "Keep running the offline node after all messages from the bag "
             "have been processed.");
 DEFINE_string(sensor_topics, "",
-              "Optional list of sensor topics for each trajectory. A single "
-              "trajectory's topics are delimited with colons, while different "
-              "trajectories are delimited with commas. If a blank list is "
-              "given for a trajectory, the default topics are used.");
+              "Optional comma-separated list of colon-separated lists of "
+              "sensor topics for each trajectory. A single trajectory's topics "
+              "are delimited with colons, while different trajectories are "
+              "delimited with commas. If a blank list is given for a "
+              "trajectory, the default topics are used. If a single "
+              "colon-separated list is given, it will be used for all "
+              "trajectories. If omitted, default topics will be used "
+              "for all trajectories.");
 
 namespace cartographer_ros {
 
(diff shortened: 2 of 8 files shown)


cartographer_ros/cartographer_ros/node.cc
```diff
@@ -118,7 +118,7 @@ Node::Node(const NodeOptions& node_options, tf2_ros::Buffer* const tf_buffer)
       &Node::PublishConstraintList, this));
 }
 
-Node::~Node() {}
+Node::~Node() { FinishAllTrajectories(); }
```
ojura: This could be added before `return`ing in  `ff (!ros::ok())`, but I thought it is more appropriate to add it here.

ojura:
@cschuet feel free to suggest how to refactor this. E.g. move `PlayableBag` and `PlayableBagMultiplexer` into `playable_bag.cc`, or `bag_management.cc` or similar.

cschuet:
@ojura https://github.com/googlecartographer/cartographer_ros/pull/643 contains the refactoring to prepare for the gRPC offline node. Should land in an hour or so.

ojura:
@cschuet Updated.

ojura:
The last commit (`Move publishing buffered back to offline node (in a callback).`) makes `PlayableBag` and `PlayableBagMultiplexer` a bit more generic and reusable, since tf republishing and buffering  is done via a callback provided in `offline_node.cc` (where it belongs). Before, we had to pass tf publisher and buffer pointers to `PlayableBag`.

ojura:
TODO: lazy adding of trajectories (when the first message from a bag is received) . For non-simultaneous (serial) bags, you get a blocked queue if you add right away the trajectories for bags which begin later.

ojura:
For now, I implemented adding trajectories lazily. As described above, the assigned trajectory IDs will now be according to trajectory times, and not the order the bags were specified.

ojura:
Okay, this should now be ready. PTAL. Let me know if you want me to split it up in multiple PRs.

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -60,19 +64,39 @@ constexpr char kTfStaticTopic[] = "/tf_static";
...
-  cartographer_ros::TrajectoryOptions trajectory_options;
-  std::tie(node_options, trajectory_options) = cartographer_ros::LoadOptions(
-      FLAGS_configuration_directory, FLAGS_configuration_basename);
+  const auto configuration_basenames =
+      cartographer_ros::SplitString(FLAGS_configuration_basenames, ',');
+  TrajectoryOptions first_trajectory_options;
+  std::tie(node_options, first_trajectory_options) =
+      LoadOptions(FLAGS_configuration_directory, configuration_basenames.at(0));
+
+  std::vector<TrajectoryOptions> trajectory_options;
```
gaschler: Remove first_trajectory_options.
Don't make == 1 a too special case.
Loop over trajectory_options with an index until bag_filenames.size().
Decide inside the loop whether to use configuration_basenames.at(0) or at(index).

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -21,37 +21,41 @@
...
-DEFINE_string(bag_filenames, "", "Comma-separated list of bags to process.");
+DEFINE_string(configuration_basenames, "",
+              "Basenames, i.e. not containing any directory prefix, of the "
+              "configuration files for each trajectory.");
 DEFINE_string(
-    urdf_filename, "",
-    "URDF file that contains static links for your sensor configuration.");
+    bag_filenames, "",
+    "Comma-separated list of bags to process. One bag per trajectory.");
+DEFINE_string(urdf_filenames, "",
```
gaschler: Is this optional? If yes, please note in the description.

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -87,9 +111,13 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
   tf2_ros::Buffer tf_buffer;
 
   std::vector<geometry_msgs::TransformStamped> urdf_transforms;
-  if (!FLAGS_urdf_filename.empty()) {
-    urdf_transforms =
-        ReadStaticTransformsFromUrdf(FLAGS_urdf_filename, &tf_buffer);
+  for (const std::string& urdf_filename :
+       cartographer_ros::SplitString(FLAGS_urdf_filenames, ',')) {
+    const auto current_urdf_transforms =
+        ReadStaticTransformsFromUrdf(urdf_filename, &tf_buffer);
```
gaschler: Won't this have name clashes by default? I am unfamiliar with this part of the code, could you please explain the new behavior?

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -132,101 +153,139 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
...
-    // always interpolate.
-    const ::ros::Duration kDelay(1.);
-    for (const rosbag::MessageInstance& msg : view) {
-      if (!::ros::ok()) {
-        break;
+    std::unordered_set<std::string> current_expected_sensor_topics;
+    if (specified_sensor_topics.at(current_bag_index).empty()) {
+      for (const std::string& topic : node.ComputeDefaultTopics(
+               trajectory_options.at(current_bag_index))) {
+        CHECK(current_expected_sensor_topics
```
gaschler: move out of if? it's the same in both cases
(create list of topics before and use that)

cartographer_ros/cartographer_ros/playable_bag.h
```diff
@@ -0,0 +1,100 @@
...
+#include "rosbag/view.h"
+#include "tf2_ros/buffer.h"
+
+namespace cartographer_ros {
+
+class PlayableBag {
+ public:
+  // Processes messages as they are about to enter the buffer. Returns a boolean
+  // indicating whether the message should enter the buffer.
+  using BufferCallback = std::function<bool(const rosbag::MessageInstance&)>;
```
gaschler: `bool /* forward_message_to_buffer */`

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -132,101 +153,139 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
...
+          }
+        }));
+  }
 
-        LOG_EVERY_N(INFO, 100000)
-            << "Processed " << (delayed_msg.getTime() - begin_time).toSec()
-            << " of " << duration_in_seconds << " bag time seconds...";
+  while (playable_bag_multiplexer.IsMessageAvailable()) {
+    const auto next_msg_tuple = playable_bag_multiplexer.GetNextMessage();
+    const rosbag::MessageInstance& msg = std::get<0>(next_msg_tuple);
```
gaschler: use std::tie?

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -132,101 +153,139 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
...
     }
+    if (msg.isType<sensor_msgs::Imu>()) {
+      node.HandleImuMessage(trajectory_id, topic,
+                            msg.instantiate<sensor_msgs::Imu>());
+    }
+    if (msg.isType<nav_msgs::Odometry>()) {
+      node.HandleOdometryMessage(trajectory_id, topic,
+                                 msg.instantiate<nav_msgs::Odometry>());
+    }
+    clock.clock = msg.getTime();
```
gaschler: add HandleNavSatFixMessage (was added after you started this PR)

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -132,101 +153,139 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
...
-        if (delayed_msg.isType<sensor_msgs::NavSatFix>()) {
-          node.HandleNavSatFixMessage(
-              trajectory_id, topic,
-              delayed_msg.instantiate<sensor_msgs::NavSatFix>());
-        }
-        clock.clock = delayed_msg.getTime();
-        clock_publisher.publish(clock);
+    playable_bag_multiplexer.AddPlayableBag(PlayableBag(
+        bag_filename, current_bag_index, ros::TIME_MIN, ros::TIME_MAX, kDelay,
+        [&](const rosbag::MessageInstance& msg) {
```
gaschler: Please explicitly write down the captures (and prefer copies if possible), I cannot see what is captured.

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -122,7 +122,7 @@ Node::Node(
...
```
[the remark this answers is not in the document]
ojura: As I explained in a review comment above:

> This could be added before returning in if (!ros::ok()), but I thought it would be more appropriate to add it here.
> Without this, if you do not finish all your trajectories, you get a failed check when exiting using Ctrl-C.

cartographer_ros/cartographer_ros/playable_bag.cc
```diff
@@ -0,0 +1,124 @@
...
+  return std::make_pair(view_->getBeginTime(), view_->getEndTime());
+}
+
+rosbag::MessageInstance PlayableBag::GetNextMessage() {
+  CHECK(IsMessageAvailable());
+  const rosbag::MessageInstance msg = buffered_messages_.front();
+  buffered_messages_.pop_front();
+  AdvanceUntilMessageAvailable();
+  if ((log_counter_++ % 10000) == 0) {
+    LOG(INFO) << "Processed " << (msg.getTime() - view_->getBeginTime()).toSec()
```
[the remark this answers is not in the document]
ojura: I thought so at first, but I think it's better to have per-bag progress indicators rather than have different bags race who is going to win the modulo 10000 condition. Let me know if you still think `LOG_EVERY_N` is better.

cartographer_ros/cartographer_ros/playable_bag.h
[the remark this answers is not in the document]
ojura: Filtering is actually a completely secondary function of this callback. The primary function is to get an early peek at tf messages and insert them into the `tf_buffer` in that lambda in `offline_node.cc`.

cartographer_ros/cartographer_ros/offline_node.cc
ojura: `ReadStaticTransformsFromUrdf` reads static transforms from an urdf and dumps them in the tf buffer. For now, you can include robot names in link names in the different urdfs, so you don't get a clash. We can open an RFC for that if you would like to change this behaviour later.

cartographer_ros/cartographer_ros/offline_node.cc
ojura: It depends on the bag. If your range data is in the tracking frame, or the transforms are in the bag, then it is optional. If this is not the case, you need to provide an urdf.

ojura:
@gaschler I am working on the PR to address your comments. In the meantime, you can try out a demo here: https://drive.google.com/open?id=11yZVQx-QPgTN9WhMv14eNtJIMlUHe8S7

Unpack it anywhere, and call `roslaunch multi_simultaneous_trajectory.launch`.

cartographer_ros/cartographer_ros/offline_node.cc
ojura: `tf_publisher` and `tf_buffer` are captured. The whole point of the lambda is not to do filtering (which I thought just looked nice), but rather to insert the transform into `tf_buffer`. This was previously achieved by having `delayed_messages`.

cartographer_ros/cartographer_ros/offline_node.cc
ojura: Can't do it with a `MessageInstance`, since you empty-construct it. Has to be a reference.

cschuet:
@ojura Sorry forgot to send my comments. If any of them have been rendered obsolete by Andre's review please ignore those.

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -132,101 +153,139 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
...
+            return true;
+          }
+        }));
+  }
 
-        LOG_EVERY_N(INFO, 100000)
-            << "Processed " << (delayed_msg.getTime() - begin_time).toSec()
-            << " of " << duration_in_seconds << " bag time seconds...";
+  while (playable_bag_multiplexer.IsMessageAvailable()) {
+    const auto next_msg_tuple = playable_bag_multiplexer.GetNextMessage();
```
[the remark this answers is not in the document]
ojura: That, and you can't empty-construct a `MessageInstance`.

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -21,37 +21,41 @@
...
 DEFINE_string(pbstream_filename, "",
               "If non-empty, filename of a pbstream to load.");
 DEFINE_bool(keep_running, false,
             "Keep running the offline node after all messages from the bag "
             "have been processed.");
+DEFINE_string(sensor_topics, "",
+              "Optional list of sensor topics for each trajectory. A single "
+              "trajectory's topics are delimited with colons, while different "
+              "trajectories are delimited with commas. If a blank list is "
+              "given for a trajectory, the default topics are used.");
```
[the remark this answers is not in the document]
ojura: Checks are there, perhaps you missed them.

cartographer_ros/cartographer_ros/node.cc
ojura: We [did](https://github.com/googlecartographer/cartographer_ros/blob/f61e513bd6479895041a9f2918505250b5bb1eb9/cartographer_ros/cartographer_ros/offline_node.cc#L229). It is and was mandatory, or you get a failed check at exiting. Previously, we would always execute that line, even if we did Ctrl-C. In this implementation, finalizing is tied to encountering the last message, which obviously will not happen when you do Ctrl-C.

It's just that I think node should handle it for us.

ojura:
Okay, updated @gaschler @cschuet. I hope I addressed everything. I hope I managed to name variables in `offline_node.cc` to be more clear.

cartographer_ros/cartographer_ros/node.cc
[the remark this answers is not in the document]
ojura: Also ptal at my answer to @cshuet: https://github.com/googlecartographer/cartographer_ros/pull/636/files#r163297172

cartographer_ros/cartographer_ros/offline_node.cc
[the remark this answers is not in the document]
ojura: AFAIK you would need [structural bindings](https://stackoverflow.com/a/43122001/1290505) from C++17 to do it.

cartographer_ros/cartographer_ros/playable_bag.h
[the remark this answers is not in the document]
gaschler: Your answer is important to understand the code.
Please rename the callback type name and variables that the code becomes self-explaining or extend the comment.

cartographer_ros/cartographer_ros/playable_bag.h
```diff
@@ -0,0 +1,96 @@
...
+#include "rosbag/view.h"
+#include "tf2_ros/buffer.h"
+
+namespace cartographer_ros {
+
+class PlayableBag {
+ public:
+  // Handles messages as they are about to enter the buffer. Returns a boolean
+  // indicating whether the message should enter the buffer.
+  using FilteringMessageHandler =
```
gaschler: Your comment on this in the code review is essential to understand the code.
Please rename the callback type name and variables that the code becomes self-explaining and/or extend this comment to fully explain the behavior.

ojura: Okay, updated. I rather documented this in `offline_node.cc` at the [call site](https://github.com/googlecartographer/cartographer_ros/pull/636/files#diff-7153329f5005d51fb75662a602a03f40R214) because that use of the handler is specific to the offline node, not `PlayableBag`. I also renamed the handler to `FilteringEarlyMessageHandler` to hint more at its intended use.

gaschler:
@wally-the-cartographer merge

ojura:
@SirVer I think the merge bot is down.

gaschler:
I'm taking care of the merge bot, @SirVer.

wally-the-cartographer:

ojura:
Great! Feel free to use the [demo](https://drive.google.com/open?id=11yZVQx-QPgTN9WhMv14eNtJIMlUHe8S7) from above to showcase this at the next open house.

cschuet:
@ojura Thanks so much for this contribution. I already made a screen recording for the next Open house of the demo bag.

gaschler:
@ojura This changed the arguments of the offline node.
Please create follow-up changes for the launch files that use this node in https://github.com/googlecartographer/cartographer_fetch https://github.com/googlecartographer/cartographer_turtlebot as needed.
(I can review them.)

PR #661, opened 2018-01-11
Enable publishing all bag messages in offline node

ojura:
This introduces an option called 'publish_all_bag_messages' which publishes everything from the bag as the offline node processes it. For example, if there is a camera on the robot, the images from the camera topic can be watched if the option is enabled, since the offline node will publish all messages from the bag (just like rosbag play).

### DIFF
--- a/cartographer_ros/cartographer_ros/offline_node.cc
+++ b/cartographer_ros/cartographer_ros/offline_node.cc
@@ -57,6 +57,9 @@ DEFINE_string(pbstream_filename, "",
 DEFINE_bool(keep_running, false,
             "Keep running the offline node after all messages from the bag "
             "have been processed.");
+DEFINE_bool(publish_all_bag_messages, false,
+            "Publish all messages from bags as they are being processed. "
+            "Transforms are controlled separately by 'use_bag_transforms'.");
 
 namespace cartographer_ros {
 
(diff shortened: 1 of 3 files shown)


ojura:
@pifon2a What happened? :-) You force pushed master to this branch.

gaschler:
@ojura Thanks for reporting. We are on it, the formatting bot make a mistake.

pifon2a:
Wally happened. This is our wonderful merge bot for git. It runs on my machine and uses my account to break stuff. :)

ojura:
@cschuet @gaschler ping for a review

cschuet:
@gaschler @ojura I agree with Andre's sentiment. I don't see this as generally useful enough to offset the cost of maintaining the code. As always thanks for putting in the effort and inventiveness to come up with new feature ideas, but I don't this is something that should go into master.

ojura:
Okay. I have no strong feelings about this - it was an one-off thing which was useful to me at some point when doing some debugging, so I tried to get it merged in hope it would be useful to someone else at some point. I see why you think maintaining this feature one is going to be difficult given the direction Cartographer development is taking (grpc and cloud slam). However, I think that the offline node _is_ trying to reimplement rosbag play in a certain kind of way.

I will rebase the other PR (the time skip option) not to be dependent on this one. Unlike this one, I think that feature is really useful, and requires minimal interventions in code.

PR #664, opened 2018-01-12
Deduplicate loading options for offline node

ojura:
This is preparation for #636. 

I noticed that there is duplicated code for loading options for the offline and GRPC offline node because they are needed while constructing the map builder for the non-GRPC offline node (and that step is the only difference between the offline node and the GRPC offline node).

I got around this by passing a map builder factory to `RunOfflineNode` instead, so we can deduplicate the code for loading options by doing it inside `RunOfflineNode`.

### DIFF
--- a/cartographer_ros/cartographer_ros/cartographer_grpc/offline_node_grpc_main.cc
+++ b/cartographer_ros/cartographer_ros/cartographer_grpc/offline_node_grpc_main.cc
@@ -15,19 +15,10 @@
  */
 
 #include "cartographer_grpc/mapping/map_builder_stub.h"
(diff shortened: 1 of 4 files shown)


ojura:
@cshuet PTAL so I can continue rebasing #636.

ojura:
@cschuet ping :)

cschuet:
@ojura Sorry again! Tomorrow! Really!

cschuet:
@wally-the-cartographer merge

cschuet:
@ojura  Thanks!

wally-the-cartographer:

PR #680, opened 2018-01-24
Add time skip option for offline node

ojura:
Introduces a "skip" option which skips first _t_ seconds.

### DIFF
--- a/cartographer_ros/cartographer_ros/offline_node.cc
+++ b/cartographer_ros/cartographer_ros/offline_node.cc
@@ -22,7 +22,6 @@
 #include <time.h>
 #include <chrono>
 
-#include "cartographer/common/optional.h"
 #include "cartographer_ros/node.h"
 #include "cartographer_ros/playable_bag.h"
 #include "cartographer_ros/split_string.h"
@@ -237,7 +236,7 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
 
   // TODO(gaschler): Warn if resolved topics are not in bags.
   std::unordered_map<int, int> bag_index_to_trajectory_id;
-  cartographer::common::optional<ros::Time> begin_time;
+  const ros::Time begin_time = playable_bag_multiplexer.PeekMessageTime();
   while (playable_bag_multiplexer.IsMessageAvailable()) {
     if (!::ros::ok()) {
       return;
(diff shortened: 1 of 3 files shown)


guilhermelawless:
@ojura this will also skip transforms in **/tf_static** when using `use_bag_transforms` and skipping time, right? Is this intented / could you consider a special case where those are included?

ojura:
Actually, tfs won’t get skipped. All messages are still processed by the main loop, just dispatching sensor messages is skipped if time < time_first + time_to_skip.

ojura:
@cschuet @gaschler PTAL, I updated this PR to remove the dependency on #661 which was closed without merging.

This is a really handy little option which enables slamming from any part of the bag.

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -60,6 +61,9 @@ DEFINE_bool(load_frozen_state, true,
 DEFINE_bool(keep_running, false,
             "Keep running the offline node after all messages from the bag "
             "have been processed.");
+DEFINE_double(skip, 0,
```
cschuet: Can we rename this to skip_seconds or anything more descriptive than skip?

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -233,12 +237,24 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
...
+    if (!::ros::ok()) {
+      return;
+    }
+
     const auto next_msg_tuple = playable_bag_multiplexer.GetNextMessage();
     const rosbag::MessageInstance& msg = std::get<0>(next_msg_tuple);
     const int bag_index = std::get<1>(next_msg_tuple);
     const bool is_last_message_in_bag = std::get<2>(next_msg_tuple);
 
+    if (!begin_time.has_value()) {
```
cschuet: Is there no way to peek the first message rather than having to do this check every time?

ojura: We can extend `PlayableBagMultiplexer` to allow for peeking the time of the next message, if you prefer.

cschuet:
@ojura Awesome! Let's do it!

cartographer_ros/cartographer_ros/offline_node.cc
cschuet: @ojura Awesome! Let's do it!

ojura:
@cschuet updated.

cschuet:
@ojura Thanks!

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #683, opened 2018-01-25
Add option to publish a pure 2D pose.

MichaelGrupp:
If the new `publish_frame_projected_to_2d` option is set to true,
the published pose will be restricted to a pure 2D pose
(no roll, pitch, or z-offset).
This prevents potentially unwanted out-of-plane poses in 2D mode
that can occur due to the pose extrapolation step (e.g. if the pose
shall be published as a 'base-footprint'-like frame).


MichaelGrupp:
This of course does not affect tilt-correction in the core algorithm, which is beneficial in any case.

However, in applications like 2D mobile robotics it can be beneficial to have a "real" 2D SLAM mode where we assume to get an SE(2) pose instead of SE(3).

cartographer_ros/cartographer_ros/trajectory_options.h
```diff
@@ -35,6 +35,7 @@ struct TrajectoryOptions {
   bool provide_odom_frame;
   bool use_odometry;
   bool use_nav_sat;
+  bool project_pose_to_2d_plane;
```
ojura: Do you think this needs to be a per-trajectory option? If not, move to node_options.h?

ojura: Also, perhaps it would be better to just publish another transform instead, to a frame e.g. `tracking_frame + "_2d"`. This transform would be obtained by embedding the pose obtained from the pose extrapolator, which would still be published normally. No need to interfere with the operation of the pose extrapolator, IMHO.

ojura: On the other hand, my proposal would make it more difficult to use the rest of the robot's TF tree with the additional `_2d` frame. I withdraw the previous comment :-)

MichaelGrupp: Yes, I also thought about just publishing an additional frame, but this would probably have unwanted side effects on the tree...

@per-trajectory option: I *personally* don't have a strong opinion on this tbh, however it sounds cleaner to me as a TrajectoryOption as the trajectory is projected into the 2D plane.

ojura: That is true, but would a scenario in which one trajectory is projected, and one isn't make sense? If not, I vote for turning it into a node option.

MichaelGrupp: For me both scenarios are fine, let's wait for another opinion :)

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -202,7 +202,13 @@ void Node::PublishTrajectoryStates(const ::ros::WallTimerEvent& timer_event) {
     const ::cartographer::common::Time now = std::max(
         FromRos(ros::Time::now()), extrapolator.GetLastExtrapolatedTime());
     stamped_transform.header.stamp = ToRos(now);
-    const Rigid3d tracking_to_local = extrapolator.ExtrapolatePose(now);
+
```
gaschler: readability: initialize non-const as it used to be.
After that, 
if (options.project) { tracking_to_local = Embed3D(tracking_to_local); }

cartographer_ros/configuration_files/backpack_2d.lua
```diff
@@ -23,6 +23,7 @@ options = {
   published_frame = "base_link",
   odom_frame = "odom",
   provide_odom_frame = true,
+  project_pose_to_2d_plane = false,
```
gaschler: Please have follow-up PRs ready for cartographer_fetch and cartographer_turtlebot.

cartographer_ros/cartographer_ros/node.cc
gaschler: readability: initialize as it used to be but without const.
After that, 
if (options.project) { tracking_to_local = Embed3D(tracking_to_local); }

cartographer_ros_msgs/msg/TrajectoryOptions.msg
```diff
@@ -18,6 +18,7 @@ string odom_frame
 bool provide_odom_frame
 bool use_odometry
 bool use_nav_sat
+bool project_pose_to_2d_plane
```
gaschler: rename to `publish_frame_projected_to_2d`, also in all lua configs and the struct

gaschler: @cschuet Do you agree with this name?

cartographer_ros/cartographer_ros/trajectory_options.h
gaschler: For consistency with the existing cartographer_ros options, TrajectoryOptions seems to fit better, even though we do not plan to change this option for each trajectory.

cartographer_ros/cartographer_ros/node.cc
MichaelGrupp: done

MichaelGrupp:
Added PR's to the `cartographer_{fetch, toyota_hsr, turtlebot}` repositories.

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #696, opened 2018-01-29
Add response to FinishTrajectory & WriteState srv.

### LINKED ISSUE #690: How do I tell if cartographer_node has finished optimising the trajectory?
The /finish_trajectory service doesn't send a response so how can I detect when the optimisations are complete? This is problem especially when working on very large maps when it can take minutes to optimise. Thank you!

MichaelGrupp:
Send a response if /finish_trajectory was successful or /write_state finished writing.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -486,7 +486,9 @@ bool Node::HandleFinishTrajectory(
     ::cartographer_ros_msgs::FinishTrajectory::Request& request,
     ::cartographer_ros_msgs::FinishTrajectory::Response& response) {
   carto::common::MutexLocker lock(&mutex_);
-  return FinishTrajectoryUnderLock(request.trajectory_id);
+  bool finished = FinishTrajectoryUnderLock(request.trajectory_id);
+  response.finished = finished;
+  return finished;
 }
 
 bool Node::HandleWriteState(
@@ -494,6 +496,7 @@ bool Node::HandleWriteState(
     ::cartographer_ros_msgs::WriteState::Response& response) {
   carto::common::MutexLocker lock(&mutex_);
   map_builder_bridge_.SerializeState(request.filename);
+  response.finished = true;
   return true;
 }
 
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -486,9 +486,9 @@ bool Node::HandleFinishTrajectory(
     ::cartographer_ros_msgs::FinishTrajectory::Request& request,
     ::cartographer_ros_msgs::FinishTrajectory::Response& response) {
   carto::common::MutexLocker lock(&mutex_);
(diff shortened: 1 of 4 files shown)


MichaelGrupp:
Adapted the description as a follow-up to the discussion in #690. A `finished == true` response does not mean that the optimizations of the map builder have passed.

The response of this PR is rather along the lines of the `PoseGraph::IsTrajectoryFinished` function ([src](https://github.com/googlecartographer/cartographer/blob/a7ed7e224f98b396762c865b81b62dc3abea2e81/cartographer/mapping_2d/pose_graph.cc#L411)). So here, `finished` means that the following steps were succesful:

* detached subscribers & sensor bridge
* all submap states of the trajectory transitioned to `SubmapState::kFinished`
* optimization dispatched

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -486,14 +486,17 @@ bool Node::HandleFinishTrajectory(
     ::cartographer_ros_msgs::FinishTrajectory::Request& request,
     ::cartographer_ros_msgs::FinishTrajectory::Response& response) {
   carto::common::MutexLocker lock(&mutex_);
-  return FinishTrajectoryUnderLock(request.trajectory_id);
+  bool success = FinishTrajectoryUnderLock(request.trajectory_id);
+  response.success = success;
+  return success;
```
ojura: This is well odd to me. Previously, we returned it directly. IIRC, when you return false here, the caller simply gets a failed service call, so you can detect it that way. I don't see the point of the `success` field, if you're still returning false when it fails. `success` is always going to be true, or else you get a failed call, IIRC again.

MichaelGrupp: (Ok, sorry for the confusion, this is going a bit off-topic from the initial motivation 😃 )

You're right, doesn't change the situation for the service caller so far. However, a `false` return value, followed by a failed service call, does not indicate for the caller whether it was a logical error (here e.g. wrong trajectory ID) or if sth in the communication failed. (see also [here](https://answers.ros.org/question/268051/why-isnt-the-response-filled-if-my-service-returns-false/?answer=268100#post-id-268100))

So my opinion now is: it looks cleaner to return always `true` to indicate a service was properly *executed*, but at the same to have additional response fields that indicate for the caller whether the service *did the right thing*. This is also true for `StartTrajectory`. The failed service calls induced by `false` return values have zero information for a caller that can't access the logs directly.

You're right about `WriteState`, I'd only like to add the new log message here.

MichaelGrupp:
Closing because the initial reason for this PR is not valid and a more general discussion (RFC) on improving services would be more beneficial (see [comment in here](https://github.com/googlecartographer/cartographer_ros/pull/696#discussion_r164734671))

cartographer_ros/cartographer_ros/node.cc
ojura: You're right in the sense that it would make differentiating between network problems and bad arguments to `FinishTrajectory` easier.

PR #704, opened 2018-02-02
Fix link for a bag with intensities.

SirVer:


wally-the-cartographer:
Merge requested by authorized user SirVer. Merge queue now has a length of 1.

PR #707, opened 2018-02-05
Offline multi-trajectory: use topic names without 'bag_n_' prefix by default

ojura:
This fixes offline_backpack_*.launch for multiple bags.


ojura:
@cschuet @gaschler So, the offline node now behaves like this:

By default, all bags will use the e.g. `scan` topic. This topic can be remapped for all bags like this: `scan:=horizontal_laser_2d` (as in `offline_backpack_2d.launch`).

Per-bag sensor remappings are still available, and will be used if specified. E.g. `bag_1_scan:=robot1/laser`.

I also added printing of resolved topics at the beginning, so it's easier to verify that we've got the remappings right.

cschuet:
@ojura Thanks for the quick PR! Since @gaschler implemented this I will defer to him. He will be back on Monday.

ojura:
Okay. Can you look into the timing issue with the bags that I mentioned in the email?

ojura:
@gaschler PTAL ping. This fixes offline_backpack_*.launch being broken for multiple sequential bags with the same configuration (i.e. topics)

ojura:
@gaschler did you have something like this in mind?

https://github.com/larics/cartographer_ros/commit/multi-bags-default-topics2

gaschler:
@ojura Yeah, this looks like a more elegant solution.
(Unless there a very common use case that it misses and I overlooked.)

ojura:
The use case where the added flexibility of the first approach would come in handy would be having different tunings for multiple bags which were recorded on the same robot (i.e. the same topic names), so you wouldn't have to specify the same topics again and again for different bags.

gaschler:
I tend to prefer the simpler solution. Thanks for all the effort.

ojura:
Okay, updated. (I also merged #736 in here just to get it to build since the current master is broken).

ojura:
@cschuet (or @gaschler) please test it out with

`roslaunch cartographer_ros offline_backpack_2d.launch bag_filenames:=/home/cschuet/Downloads/b0-2014-08-14-13-23-01.bag,/home/cschuet/Downloads/b0-2014-08-14-13-36-48.bag`

ojura:
@gaschler Also, if you have a single configuration file, but multiple bags with different topics, you will be forced to repeat the configuration filename multiple times, since you now won‘t be able to specify per-bag remappings otherwise. We should probably document this all somewhere, can you please handle that?

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -152,8 +152,21 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
       },
       false /* oneshot */, false /* autostart */);
 
-  auto bag_expected_sensor_ids =
-      node.ComputeDefaultSensorIdsForMultipleBags(bag_trajectory_options);
+  std::vector<
+      std::set<cartographer::mapping::TrajectoryBuilderInterface::SensorId>>
+      bag_expected_sensor_ids;
+  if (configuration_basenames.size() == 1) {
+    bag_expected_sensor_ids = node.ComputeDefaultSensorIdsForMultipleBags(
```
gaschler: more readable: `auto expected_sensor_ids =` and then below you can use the n-elements vector constructor instead of insert.

ojura:
@gaschler Ok, updated to avoid insert.

gaschler:
@wally-the-cartographer merge

ojura:
@gaschler Wally seems down.

wally-the-cartographer:
Merge requested by authorized user gaschler. Merge queue now has a length of 2.

PR #710, opened 2018-02-07
Unfrozen trajectories

ojura:
Replaces #517.

Follow up to googlecartographer/cartographer#891.


cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -105,10 +105,16 @@ MapBuilderBridge::MapBuilderBridge(
...
-  LOG(INFO) << "Loading map '" << map_filename << "'...";
-  cartographer::io::ProtoStreamReader stream(map_filename);
-  map_builder_->LoadMap(&stream);
+void MapBuilderBridge::LoadState(const std::string& state_filename,
+                                 bool load_frozen_state) {
+  // Check if suffix of the state file is ".pbstream".
+  const std::string suffix = ".pbstream";
+  CHECK_EQ(state_filename.substr(
+               std::max<int>(state_filename.size() - suffix.size(), 0)),
+           suffix);
```
cschuet: Can you add a message here along the lines of "Must be a .pbstream file"?

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -309,9 +313,21 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
...
 
   if (::ros::ok()) {
-    const std::string output_filename = bag_filenames.front() + ".pbstream";
-    LOG(INFO) << "Writing state to '" << output_filename << "'...";
-    node.SerializeState(output_filename);
+    std::string output_filename;
+    const std::string suffix = ".pbstream";
+    if (bag_filenames.size() > 0) {
+      output_filename = bag_filenames.front();
+    } else {
```
cschuet: It seems to me that when you load N .pbstreams in unfrozen state, they are not optimized relative to one another (no inter pbstream constraints are searched for). Given that why would anyone want to write a pbstream in a situation where no bag file has been played?

ojura: It's useful to be able to launch an offline node instance which only serves a loaded unfrozen `.pbstream` for visualization (using `keep_running = true`). However, you're right about further serialization making no sense in that scenario. I'll change it to avoid serializing when no bags are played.

cschuet: OK, SG. Having N .pbstream of the same are and being able to fire up offline node to optimize them with respect to one another might be useful though. So maybe we add that in the future. But for now agreed remove.

ojura: Sounds like an interesting idea. Maybe it belongs in an RFC :)

cartographer_ros/cartographer_ros/node_main.cc
```diff
@@ -28,12 +28,16 @@ DEFINE_string(configuration_directory, "",
 DEFINE_string(configuration_basename, "",
               "Basename, i.e. not containing any directory prefix, of the "
               "configuration file.");
-DEFINE_string(map_filename, "", "If non-empty, filename of a map to load.");
+DEFINE_string(load_state_filename, "",
```
gaschler: @ojura, please make the follow-up change in cartographer_fetch/demo_freight_localization.launch

PR #713, opened 2018-02-07
Publish Landmark markers for RViz.

pifon2a:
[RFC=0011](https://github.com/googlecartographer/rfcs/blob/master/text/0011-landmarks.md)

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -25,8 +25,9 @@
 #include "cartographer_ros_msgs/StatusResponse.h"
 
 namespace cartographer_ros {
-
 namespace {
+
+using ::cartographer::transform::Rigid3d;
 
 constexpr double kTrajectoryLineStripMarkerScale = 0.07;
 constexpr double kConstraintMarkerScale = 0.025;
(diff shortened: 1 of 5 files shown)


cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -55,6 +56,32 @@ visualization_msgs::Marker CreateTrajectoryMarker(const int trajectory_id,
   return marker;
 }
 
+int GetLandmarkIndex(const std::string& landmark_id) {
+  static auto* landmark_id_to_index = new std::unordered_map<std::string, int>;
```
cschuet: Ehm...

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -55,6 +56,32 @@ visualization_msgs::Marker CreateTrajectoryMarker(const int trajectory_id,
...
+  }
+  return it->second;
+}
+
+visualization_msgs::Marker CreateLandmarkMarker(const std::string& landmark_id,
+                                                const Rigid3d& landmark_pose,
+                                                const std::string& frame_id) {
+  visualization_msgs::Marker marker;
+  marker.ns = "Landmarks";
+  int landmark_index = GetLandmarkIndex(landmark_id);
```
cschuet: Why not just hash the landmark_id?

pifon2a: Hashes can have collisions.

ojura: I think that using a string for an ID doesn't mesh well with the rest of the cartographer codebase - all other IDs (submap, node) are integer-based.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -55,6 +56,32 @@ visualization_msgs::Marker CreateTrajectoryMarker(const int trajectory_id,
   return marker;
 }
 
+int GetLandmarkIndex(
+    const std::string& landmark_id,
+    std::unordered_map<std::string, int>* landmark_id_to_index) {
+  auto it = landmark_id_to_index->find(landmark_id);
```
MichaelGrupp: Why not just use

```c++
const int new_index = landmark_id_to_index->size()++;
landmark_id_to_index->emplace(std::make_pair(landmark_id, new_index));
return landmark_id_to_index->at(landmark_index);
```
Should get rid of the if condition.
http://en.cppreference.com/w/cpp/container/unordered_map/emplace

(not tested)

pifon2a: I think it is less readable.

PR #717, opened 2018-02-09
RViz settings for landmarks.

pifon2a:
[RFC=0011](https://github.com/googlecartographer/rfcs/blob/master/text/0011-landmarks.md)


pifon2a:
[RFC=0011](https://github.com/googlecartographer/rfcs/blob/master/text/0011-landmarks.md)

PR #732, opened 2018-02-22
Add a ROS message for landmark observations.

pifon2a:
[RFC](https://github.com/googlecartographer/rfcs/blob/master/text/0011-landmarks.md)


cartographer_ros_msgs/msg/LandmarkEntry.msg
```diff
@@ -0,0 +1,18 @@
+ 2018 The Cartographer Authors
```
cschuet: #?

PR #738, opened 2018-02-26
Add Magazino public datasets to data.rst

MichaelGrupp:


MichaelGrupp:
As discussed.

cschuet:
@wally-the-cartographer merge

SirVer:
2c: Having it in the repo means that for all eternity everybody needs to download 200MB of history when checking out the magazino repo. I suggest [purging the large files from git history](https://stackoverflow.com/questions/2100907/how-to-remove-delete-a-large-file-from-commit-history-in-git-repository) and instead serve it from google cloud storage.


@cschuet @MichaelGrupp Can this data instead be downloaded and put to the google cloud store buckets by Google?

MichaelGrupp:
That's why we use the Git LFS extension: https://git-lfs.github.com/ I.e. versioning pointers to a storage instead of big binary blobs in the repo. Cloning it is cheap.

You can put in a Google storage if you prefer that for your purposes.

SirVer:
I see, thanks for explaining. I remember that we looked at lfs when we decided to open source cartographer's public data and decided that it's additional knowledge did not make it a great fit for never changing data. I have no stake anymore in this, but as a user I'd prefer having all data in the same place.

MichaelGrupp:
@cschuet What's your opinion? In any case I'd keep the LFS in our repo, for the same reason (everything in one place). Regarding the storage in the public dataset list, we can go for what fits best for you.

By the way: if we keep the LFS files in there, should I change the links to direct ones? (they currently link to the GitHub page)

cschuet:
@MichaelGrupp This is ultimately up to you. If you are ok with it, I could move them into our public-datasets Google Cloud Storage bucket.

MichaelGrupp:
Both options are fine for me - if there's no pressing need from your side, then let's leave it like it is.

Otherwise, feel free to copy them to a Google Storage if this is more convenient for you.

cschuet:
@MichaelGrupp I was just working on the continuous quality control pipeline and hosting them in a GCS bucket makes the integration easier. I copied the datasets to our public data bucket and updated the links.

cschuet:
@wally-the-cartographer merge

PR #739, opened 2018-02-27
Add Fetch and Toru platform to Jenkins build

cschuet:


cschuet:
@wally-the-cartographer merge

PR #745, opened 2018-02-28
Fix pbstream_map_publisher (follow #712)

MichaelGrupp:
Applies the proto deserialization changes that
were introduced in PR #712.


MichaelGrupp:
Without this, it produces segfaults.

@cschuet  I guess this was accidentally missed because the pbstream_map_publisher was merged one commit after these changes in #712.

Maybe we could have some small smoke tests that check if "main" executables actually launch with minimal test data?

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #768, opened 2018-03-13
Use gmock after catkin

gaschler:
https://answers.ros.org/question/199680/catkin-doesnt-play-nice-with-google-mock/?answer=221147

### DIFF
--- a/cartographer_ros/CMakeLists.txt
+++ b/cartographer_ros/CMakeLists.txt
@@ -40,9 +40,14 @@ include("${CARTOGRAPHER_CMAKE_DIR}/functions.cmake")
 set(BUILD_SHARED_LIBS OFF)
 option(BUILD_GRPC "build features that require Cartographer gRPC support" false)
 google_initialize_cartographer_project()
-google_enable_testing()
 
 find_package(catkin REQUIRED COMPONENTS ${PACKAGE_DEPENDENCIES})
+
+google_enable_testing()
+# Override Catkin's GTest configuration to use GMock.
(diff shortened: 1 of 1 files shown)


cartographer_ros/CMakeLists.txt
```diff
@@ -40,10 +40,15 @@ include("${CARTOGRAPHER_CMAKE_DIR}/functions.cmake")
...
 google_initialize_cartographer_project()
-google_enable_testing()
 
 find_package(catkin REQUIRED COMPONENTS ${PACKAGE_DEPENDENCIES})
 
+google_enable_testing()
+# Override Catkin's GTest configuration to use GMock.
+set(GTEST_FOUND TRUE)
+set(GTEST_INCLUDE_DIRS ${GMOCK_INCLUDE_DIRS})
+set(GTEST_LIBRARIES ${GMOCK_LIBRARIES})
```
dirk-thomas: I am not sure if this is still necessary. The latest version of `catkin` already does use GMock when it is available.

ojura:
@gaschler this is not a good fix for #776. I submitted a proper fix, https://github.com/ros/catkin/pull/927, and it has been merged. I asked if they can make a release so we can get cartographer_ros CI unstuck.

gaschler:
@ojura, I'm on it. My latest plan is to use cmake's `ExternalProject_Add` to build it reproducibly. We cannot wait for upstream changes.
(Abandoning this PR.)

ojura:
Maybe consider just applying that patch in the Docker instance?

PR #777, opened 2018-03-22
Avoid failed CHECK when running offline node with no bags.

ojura:
Bug introduced in #680.


ojura:
@cschuet would have been better to use [an optional](https://github.com/googlecartographer/cartographer_ros/pull/680#discussion_r172270902) :)

cschuet:
@ojura Why would you want to run the offline node w/o bags?

ojura:
We‘ve already had this discussion in #710 :-)
https://github.com/googlecartographer/cartographer_ros/pull/710#comment-170626905

PR #787, opened 2018-03-27
Pass GitHub access token to Dockerfile.

MichaelGrupp:
See here: https://github.com/magazino/cartographer_magazino/issues/9

Fixes https://github.com/googlecartographer/cartographer/issues/1012


MichaelGrupp:
Please add a token to your Travis settings as explained in the issue in `cartographer_magazino` repository.

ojura:
Do the fork owners have to set up anything in their Travis settings in order to get this to work?

MichaelGrupp:
I think so, if you run your own Travis pipeline with your fork. But adding a environment variable `GITHUB_TOKEN` to the settings should be enough.

As far as I understood any GitHub token should be fine as it is only for authentication without access privileges.

MichaelGrupp:
We at Magazino are running our *actual* forks on an internal Jenkins server, so you would need to test it yourself.

ojura:
Using Travis with a Github fork of Cartographer is a single-click experience, it would be cool if it could remain such and still be able to use this :)

MichaelGrupp:
Then I guess the only option would be to add the token environment variable in plain text to `travis.yml`.

Or what do you think?

Also: @gaschler PTAL

MichaelGrupp:
Small update to my previous comment:

If `GITHUB_TOKEN` is not specified on Travis by a fork maintainer, it will just pass as empty. Then the call to the GitHub API will still be possible, just that it will go unauthenticated as now. 

(That's also the reason why the build for this pull request succeeded without a token on Travis yet).

MichaelGrupp:
Update nr. 2: see my latest commit, I added a check for the environment variable at the start of a build. This way it explicitly fails if there's no token, see my test fork: https://api.travis-ci.org/v3/job/359274845/log.txt

Interesting enough, the Google build seems to have the `GITHUB_TOKEN` already because it doesn't fail.

cschuet:
@MichaelGrupp Thanks! Interestingly we don't have a GITHUB_TOKEN env variable defined for any of our Travis projects AFAICT. So I am also wondering why you are not running into the check. Can you echo the token somewhere so we can see it in the logs?

ojura:
Good job on adding a hard CI error to point this out to fork owners, I was thinking of suggesting this.

MichaelGrupp:
Uhm... I added a temporary commit with echo.

If you have a look at the build of my follow-up in `cartographer_turtlebot` for example, the error happens: https://travis-ci.org/googlecartographer/cartographer_turtlebot/jobs/359280493

MichaelGrupp:
Now it failed as expected, no token is present. I have no idea why it worked the first time (??).

MichaelGrupp:
I just noticed "we do not provide these values to untrusted builds, triggered by pull requests from another repository" - https://docs.travis-ci.com/user/environment-variables/
This would not solve the problem.

Would you be fine with providing a plain text token in .travis.yml?

SirVer:
Fwiw, you can put the github token into the environment in travis and pass it to docker that way. All forks would use the same setting, keeping travis a possibility: https://docs.travis-ci.com/user/environment-variables/#Defining-encrypted-variables-in-.travis.yml

SirVer:
Never mind, they are actually also not provided to untrusted builds.

MichaelGrupp:
We now use an unencrypted environment variable in Travis env settings. The unencrypted part is the trick to make it work with pull requests.

BTW, maybe I should add a comment or log message for fork owners that explains this.

PR #811, opened 2018-04-06
Take frozen state into account when finishing trajectories.

MichaelGrupp:
Until now, the error response of an /finish_trajectory request for a
frozen trajectory was 'Trajectory ... is not created yet.'.

This is a lie. The new response is more accurate because the trajectory
__is__ created, but it just can't be finished because it's frozen.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -443,6 +443,16 @@ bool Node::ValidateTopicNames(
 cartographer_ros_msgs::StatusResponse Node::FinishTrajectoryUnderLock(
     const int trajectory_id) {
   cartographer_ros_msgs::StatusResponse status_response;
+
+  // First, check if we can actually finish the trajectory.
+  if (map_builder_bridge_.GetFrozenTrajectoryIds().count(trajectory_id)) {
+    const std::string error =
+        "Trajectory " + std::to_string(trajectory_id) + " is frozen.";
+    LOG(INFO) << error;
+    status_response.code = cartographer_ros_msgs::StatusCode::INVALID_ARGUMENT;
+    status_response.message = error;
+    return status_response;
+  }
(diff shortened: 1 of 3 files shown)


cartographer_ros/cartographer_ros/node.cc
```diff
@@ -443,6 +443,16 @@ bool Node::ValidateTopicNames(
 cartographer_ros_msgs::StatusResponse Node::FinishTrajectoryUnderLock(
     const int trajectory_id) {
   cartographer_ros_msgs::StatusResponse status_response;
+
+  // First, check if we can actually finish the trajectory.
+  if (map_builder_bridge_.GetFrozenTrajectoryIds().count(trajectory_id)) {
+    const std::string error =
+        "Trajectory " + std::to_string(trajectory_id) + " is frozen.";
+    LOG(INFO) << error;
```
ojura: Nit: LOG(INFO) << "Error string"?

MichaelGrupp: Should be LOG(ERROR), changed.

ojura: Ah, didn't catch that. I was referring to creating a string variable - you can just stream the error message directly, without creating a variable.

MichaelGrupp: Oh yes, also true :+1:

MichaelGrupp: However, it's also true that the error string gets both logged and passed to the response, so the variable has a right to stay ;)

ojura: Sorry, I missed that :)

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user cschuet. Merge queue now has a length of 2.

PR #812, opened 2018-04-06
Fix race-condition when attempting to fetch trimmed submap slices.

MichaelGrupp:
A simple solution for a slightly more complex scenario:
- a pure localization trajectory `X` gets finished & trimmed in the main node
- at the same time, the occupancy_grid_node handles an outdated SubmapList
  message in which a submap ID `id` of trajectory `X` is still present
- the call to FetchSubmapTextures(`id`, ...) leads to a crash

With this fix, the trimmed submap IDs are just ignored until the next
iteration (in which the occupancy grid node removes the trimmed IDs).

### DIFF
--- a/cartographer_ros/cartographer_ros/submap.cc
+++ b/cartographer_ros/cartographer_ros/submap.cc
@@ -20,6 +20,7 @@
 #include "cartographer/common/port.h"
 #include "cartographer/transform/transform.h"
 #include "cartographer_ros/msg_conversion.h"
+#include "cartographer_ros_msgs/StatusCode.h"
 #include "cartographer_ros_msgs/SubmapQuery.h"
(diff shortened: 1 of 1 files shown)


MichaelGrupp:
Noticed this while playing around with a new initial pose handler that finishes the current trajectory and starts a new one for each initial pose message from Rviz. I could reproduce this crash several times.

Please note that the error message `Requested submap 1 from trajectory 10 but it does not exist: maybe it has been trimmed.` is still logged once with the fix.

cschuet:
@wally-the-cartographer merge

cschuet:
@MichaelGrupp Thanks!

wally-the-cartographer:

Issue #780, opened 2018-03-23
Use catkin build instead of catkin_make_isolated


ojura:
The build tool `catkin build` (in package python-catkin-tools) is a more modern version of `catkin_make`/`catkin_make_isolated`. It builds packages in isolation like `catkin_make_isolated`, while having much nicer screen output and CLI.

Change documentation to use `catkin build`?

PR #813, opened 2018-04-06
Ros integration test

gaschler:
Adds a rostest that brings up the offline node, subscribes to tf results and verifies some motion was recognized.


gaschler:
Filed upstream ROS question: https://answers.ros.org/question/288290/catkin_make_isolated-run_tests/

gaschler:
@mikaelarguedas, could you please take a look at https://answers.ros.org/question/288290/catkin_make_isolated-run_tests/, I'd really appreciate your help because I got stuck here.

ojura:
@gaschler #780?

gaschler:
@ojura, `catkin build` is certainly a good idea, but dropping `catkin_make_isolated` support, which is our main build system, is not an easy decision I would push for to make a single integration test work. (If we were to switch build systems, there should be enough time where both are maintained, so it is quite an effort.)

mikaelarguedas:
@gaschler I had a quick look at this. I can successfully reproduce the error with ninja and not make.

What seems to be the issue is the generated command for the target `run_tests` is that the generated ninja build file has a trailing backslash. When using make the backslash is present as well but doesnt cause any issue.

I didn't find anywhere in catkin's CMake code where these backslashes are added or how to prevent this from failing, but hopefully that'll give you a lead to explore. I tested only on the example you provided at `https://github.com/gocarlos/ros_unit_tests_example`. I'd be curious to know if that fixes the issue reported here as well.

in my case the command looks like this:
```
  COMMAND = cd /root/test_ws/build_isolated/ros_unit_tests_example && /root/test_ws/build_isolated/ros_unit_tests_example/catkin_generated/env_cached.sh /usr/bin/python /opt/ros/lunar/share/catkin/cmake/test/run_tests.py /root/test_ws/build_isolated/ros_unit_tests_example/test_results/ros_unit_tests_example/rostest-launch_node_launcher.xml /opt/ros/lunar/share/rostest/cmake/../../../bin/rostest\ --pkgdir=/root/test_ws/src/ros_unit_tests_example\ --package=ros_unit_tests_example\ --results-filename\ launch_node_launcher.xml\ --results-base-dir\ "/root/test_ws/build_isolated/ros_unit_tests_example/test_results"\ /root/test_ws/src/ros_unit_tests_example/launch/node_launcher.launch\
```
if I remove the backslash (or add a whitespace before the backslash) on the last line I can run the test with ninja as well as make:
```
==> ninja build.ninja in '/root/test_ws/build_isolated/ros_unit_tests_example'
ninja: no work to do.
==> ninja run_tests -j12 -l12 in '/root/test_ws/build_isolated/ros_unit_tests_example'
[1/6] cd /root/test_ws/build_isolated/ros_unit_tests_example && /usr/bin/pyt...ws/build_isolated/ros_unit_tests_example/test_results/ros_unit_tests_example
Removing test result files from '/root/test_ws/build_isolated/ros_unit_tests_example/test_results/ros_unit_tests_example'
[6/6] cd /root/test_ws/build_isolated/ros_unit_tests_example && /root/test_w...sults"\ /root/test_ws/src/ros_unit_tests_example/launch/node_launcher.launch
... logging to /root/.ros/log/rostest-e2d4008994f8-6142.log
[ROSUNIT] Outputting test results to /root/test_ws/build_isolated/ros_unit_tests_example/test_results/ros_unit_tests_example/rostest-launch_node_launcher.xml
[Testcase: testmy_unit_test] ... ok

[ROSTEST]-----------------------------------------------------------------------

[ros_unit_tests_example.rosunit-my_unit_test/test_case_1][passed]

SUMMARY
 * RESULT: SUCCESS
 * TESTS: 1
 * ERRORS: 0
 * FAILURES: 0
```

gaschler:
@mikaelarguedas, thank you for investigating this.
I can confirm that manually removing the backslash at the end of two command lines in `build.ninja` makes the rostest pass. It does get stuck later after "verify result" but I'd assume this would be a different issue on our side.

Here are the logs when I execute 
```
CTEST_OUTPUT_ON_FAILURE=1  catkin_make_isolated --use-ninja --install 
```
manually remove the two slashes, and then
```
CTEST_OUTPUT_ON_FAILURE=1  catkin_make_isolated --use-ninja --install --catkin-make-args run_tests
```

```

[5/6] cd ~/workspace/catki...pher_ros/cartographer_ros/tests/offline_backpack_2d.test
WARNING: Logging before InitGoogleLogging() is written to STDERR
I0423 16:50:46.152065 228357 result_subscriber_test.cc:64] Observed simulation time:5.073805400
I0423 16:50:46.152554 228357 result_subscriber_test.cc:65] Observed travel distance:0.581736
... logging to ~/.ros/log/rostest-228325.log
[ROSUNIT] Outputting test results to ~/workspace/catkin_ws/build_isolated/cartographer_ros/test_results/cartographer_ros/rostest-tests_offline_backpack_2d.xml
[Testcase: testsubscribe_moving_base_link] ... ok

[ROSTEST]-----------------------------------------------------------------------

[cartographer_ros.rosunit-subscribe_moving_base_link/ReceiveMovingBaseLink][passed]

SUMMARY
 * RESULT: SUCCESS
 * TESTS: 1
 * ERRORS: 0
 * FAILURES: 0

rostest log file is in ~/.ros/log/rostest-228325.log
-- run_tests.py: execute commands
  /opt/ros/lunar/share/rostest/cmake/../../../bin/rostest --pkgdir=~/workspace/catkin_ws/src/cartographer_ros/cartographer_ros --package=cartographer_ros --results-filename tests_offline_backpack_2d.xml --results-base-dir ~/workspace/catkin_ws/build_isolated/cartographer_ros/test_results ~/workspace/catkin_ws/src/cartographer_ros/cartographer_ros/tests/offline_backpack_2d.test
-- run_tests.py: verify result "~/workspace/catkin_ws/build_isolated/cartographer_ros/test_results/cartographer_ros/rostest-tests_offline_backpack_2d.xml"

```

(After this line, it freezes. `ps` reports that the processes `catkin_make_isolated`, `python`, and my `result_subscriber_node` that is part of the test are still running.)

Interesting. Are there any cmake flags related to the paths where the backslash could come from, which are incompatible with the ninja generator? Any other work-around than manually removing the backslashes?

mikaelarguedas:
> Are there any cmake flags related to the paths where the backslash could come from, which are incompatible with the ninja generator?

@gaschler TBH I'm not very familiar with ninja and it's not clear to me why the backslash is not processed properly there.
Preventing CMake from adding magic seems to fix it (if you can give a shot to this commit https://github.com/mikaelarguedas/catkin/commit/ec5bf37c9e9bcaa20e991c931b81046e26885043 that seems to fix it for me).
But I would need to look into more details what the implications of this change are before considering contributing it to catkin. I may open a PR tomorrow just to get feedback from people that are more versed in that matter.

ojura:
[Isn't it](https://cmake.org/cmake/help/v3.0/command/add_custom_target.html) always a good idea to use `VERBATIM`?

> If `VERBATIM` is given then all arguments to the commands will be escaped properly for the build tool so that the invoked command receives each argument unchanged. Note that one level of escapes is still used by the CMake language processor before `add_custom_target even` sees the arguments. Use of `VERBATIM` is recommended as it enables correct behavior. When `VERBATIM` is not given the behavior is platform specific because there is no protection of tool-specific special characters.

mikaelarguedas:
> Isn't it always a good idea to use VERBATIM?

That's my reading of the docs as well, though as the option has been in CMake since the early days and is still not the default, I suspect there is a reason for it.

Were you able to confirm that this PR passes with the `VERBATIM` fix ?

mikaelarguedas:
PR opened https://github.com/ros/catkin/pull/935

gaschler:
@mikaelarguedas 
Thanks for digging deeper and proposing a fix.
Reproducing took longer than expected because I had to make an additional change:
For me, the rostest passes after adding a third `VERBATIM`, to the target `_run_tests_${PROJECT_NAME}_${type}_${name}`.

```
# hidden test target which depends on building all tests and cleaning test results
  add_custom_target(_run_tests_${PROJECT_NAME}_${type}_${name}
    COMMAND ${cmd}
    VERBATIM)
```
I use Ros Lunar with Catkin 0.7.11 and Ninja 1.8.2.

Thanks for preparing the PR, hope the fix (with three verbatims) will be approved.

mikaelarguedas:
> I use Ros Lunar with Catkin 0.7.11 and Ninja 1.8.2.

Do you think it's possible to modify the dockerfiles temporarily to test the fix on all distros ? This way we make sure that it solves the problem for your entire CI once released.

> Thanks for preparing the PR, hope the fix (with three verbatims) will be approved.

I'll add the third verbatim to the PR (we may need to add it to other places as well but that's harder to figure out without reproducible failures, it may work to add it everywhere but I'd like to here back from the maintainer before).
We need to keep in mind that even if it's approved and merged, it will likely be a while to get this fixed released as releasing catkin means rebuilding every ROS package.
We'll also need to see if it's approved for backport to all ROS distros as well

cschuet:
@gaschler Should we close this for now?

mikaelarguedas:
@gaschler @cschuet Quick update on the catkin fix:

- It has been merged in the branch targetting ROS Kinetic, Lunar and Melodic.
- It will likely not be back ported to Indigo
- It has been released in melodic and will be released in Lunar and Kinetic next time catkin is released in these distributions

gaschler:
Re-opening to re-run build bot

PR #821, opened 2018-04-10
Internal cleanup.

zjwoody:
Fix lint error.


zjwoody:
I see, 
how about change all the header guards with double "CARTOGRAPHER_ROS", since the files are all under path cartographer_ros/cartographer_ros/... ?

gaschler:
@cschuet, what do you think about making all include guards longer? I think this is low priority but I'm not opposed.

cschuet:
@zjwoody @gaschler Why is this change necessary? Why don't the other header files generate a linter error?

zjwoody:
The third party header guard is not google style, if it's length > 80 characters it won't pass is_header_guard check, so I have to change it to google style for long header guard. Most of the files don't have long header guard, so they won't generate the lint error. 
I think it will be more consistency to change all the header guard by the path which obey the google style.

zjwoody:
Hi @gaschler , I modified all the header guards in cartographer_ros, thanks for the review!

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #822, opened 2018-04-11
Update ogre_slice.h

XuHaiwang:
Add three head files. Fix the error during compiling in Arch Linux.


cschuet:
@XuHaiwang Thanks for the PR. Is it not sufficient to add these headers to the .cc?

cschuet:
@XuHaiwang Friendly ping.

XuHaiwang:
hi, it is better that these headers exist in ogre_slice.h because of the GCC version in Arch Linux.

ojura:
Can you add more details? Which GCC version, what symbols are used in ogre_slice.h which are found in these headers?

PR #834, opened 2018-04-20
Getter for trajectories in AssetsWriter

schwoere:
Added a getter for the trajectories of the pose graph to the AssetsWriter and moved CreateFileWriterFactory outside of the AssetsWriter class.

To register new points processors, PointsProcessors need to be constructed from outside the AssetsWriter. This may require the trajectories of the pose graph. Therefore, the getter for the trajectories was added to the AssetsWriter.

The static class method CreateFileWriterFactory was moved outside the AssetsWriter to communicate its independence of class internals of the AssetsWriter.

### DIFF
--- a/cartographer_ros/cartographer_ros/assets_writer.h
+++ b/cartographer_ros/cartographer_ros/assets_writer.h
@@ -46,9 +46,11 @@ class AssetsWriter {
            const std::string& configuration_basename,
            const std::string& urdf_filename, bool use_bag_transforms);
 
-  // Creates a FileWriterFactory which creates a FileWriter for storing assets.
-  static ::cartographer::io::FileWriterFactory CreateFileWriterFactory(
-      const std::string& file_path);
+  // Returns all trajectories of the pose graph.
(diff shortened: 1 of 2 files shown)


cartographer_ros/cartographer_ros/assets_writer.h
```diff
@@ -58,6 +60,10 @@ class AssetsWriter {
       point_pipeline_builder_;
 };
 
+// Creates a FileWriterFactory which creates a FileWriter for storing assets.
+::cartographer::io::FileWriterFactory CreateFileWriterFactory(
```
wohe: Why was that changed? Seemed perfectly fine to me.

schwoere: Okay, I thought it helps to communicate the independence from AssetsWriter class internals. I reverted back.

cschuet:
@schwoere Can we close this for now? I don't see any activity on this.

schwoere:
@cschuet Can be closed, yes. Thanks!

PR #874, opened 2018-05-24
Assets writer (ROS map) urdf typo fix

rsmallegoor:
The ROS map assets writer launch file can now find the default urdf file when no argument is provided.


rsmallegoor:
#875

PR #891, opened 2018-06-06
Fix memory leak in simulations by removing wall timers.

MichaelGrupp:
Fixes the problem of ever-growing memory after `rosbag play --clock` finishes,
as discussed in https://github.com/googlecartographer/cartographer/issues/1182

The wall timers caused the timer callback that publishes TF data to be called
even if no simulated `/clock` was published anymore.
As the TF buffer cache time of the TF listener seems to be based on
the ROS time instead of wall clock, it could grow out of bounds.

Now, `ros::Timer` plays nicely with both normal (wall) and simulated time and
no callbacks are executed if `/clock` stops in simulation.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -126,9 +126,9 @@ Node::Node(
   timers_.push_back(node_handle_.createTimer(
       ::ros::Duration(node_options_.trajectory_publish_period_sec),
       &Node::PublishLandmarkPosesList, this));
-  timers_.push_back(node_handle_.createTimer(
-      ::ros::Duration(kConstraintPublishPeriodSec),
-      &Node::PublishConstraintList, this));
+  timers_.push_back(
+      node_handle_.createTimer(::ros::Duration(kConstraintPublishPeriodSec),
+                               &Node::PublishConstraintList, this));
 }
 
 Node::~Node() { FinishAllTrajectories(); }
@@ -278,8 +278,7 @@ void Node::PublishLandmarkPosesList(
   }
 }
 
-void Node::PublishConstraintList(
-    const ::ros::TimerEvent& unused_timer_event) {
+void Node::PublishConstraintList(const ::ros::TimerEvent& unused_timer_event) {
   if (constraint_list_publisher_.getNumSubscribers() > 0) {
     carto::common::MutexLocker lock(&mutex_);
     constraint_list_publisher_.publish(map_builder_bridge_.GetConstraintList());
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -114,20 +114,20 @@ Node::Node(
       node_handle_.advertise<sensor_msgs::PointCloud2>(
           kScanMatchedPointCloudTopic, kLatestOnlyPublisherQueueSize);
 
-  wall_timers_.push_back(node_handle_.createWallTimer(
-      ::ros::WallDuration(node_options_.submap_publish_period_sec),
+  timers_.push_back(node_handle_.createTimer(
+      ::ros::Duration(node_options_.submap_publish_period_sec),
       &Node::PublishSubmapList, this));
-  wall_timers_.push_back(node_handle_.createWallTimer(
-      ::ros::WallDuration(node_options_.pose_publish_period_sec),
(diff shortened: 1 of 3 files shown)


MichaelGrupp:
By the way, I first thought that `rosbag play --clock --keep-alive` fixes the problem.... it fixes the memory problem, but would also keep pose extrapolation alive, leading to unwanted motion after the bagfile finishes.

gaschler:
Cool thing this resolved the tf data memory accumulation.
Generally, this sounds fine for the online node with and without simulated time.

Concerning the lua parameters, the publish_period would change in meaning and refer to the simulated clock in the simulated case. 

One thing I am not sure about: Will this change also play nicely with the offline node? The offline node has its own clock_publisher. Have you thought about this and tested it?

MichaelGrupp:
The offline node works also with this patch. The publish period of the topics will adapt to the clock of the offline node as with any other simulated clock.

pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:

ojura:
Hmmm. I am not a huge fan of this, I would like the submap/trajectory node list to keep being published when the offline node enters its `keep_waiting` stage, when the simulation clock is standing still.

Could we not just stop the timer which publishes the transforms (`PublishTrajectoryStates`)? All trajectories are finished at that point, we should not be publishing their transforms. This would fix bloating of the tf buffer.

ojura:
> Also, shutting down the subscribers of the node, i.e. calling /finish_trajectory after rosbag play finishes, stops the memory accumulation.

You should absolutely finish your trajectories after you played them to their end, which will stop publishing their transforms and in turn stop bloating the tf buffer. So this was partly the user's fault. 
Changing the timer for `PublishTrajectoryStates` not to be a `WallTimer` to prevent memory bloating if the user forgets to finish their trajectories seems reasonable, but I would prefer if we did not have unnecessarily made this change to other timers.

ojura:
> Also, shutting down the subscribers of the node, i.e. calling /finish_trajectory after rosbag play finishes, stops the memory accumulation.

You should absolutely finish your trajectories after you played them to their end, which will stop publishing their transforms and in turn stop bloating the tf buffer. So this was partly the user's fault. 
Changing the timer for `PublishTrajectoryStates` not to be a `WallTimer` to prevent memory bloating if the user forgets to finish their trajectories seems reasonable, but I would prefer if we were not making his change to other timers.

MichaelGrupp:
> You should absolutely finish your trajectories after you played them to their end, which will stop publishing their transforms and in turn stop bloating the tf buffer. So this was partly the user's fault.

This only accounts for batch processing. If you run with simulated time and e.g. pause the data publishing for a while, the memory shouldn't leak. This is not a user problem, also in the first case where finishing trajectories could make sense it's totally non-obvious why the memory increases unless you know that a wall timer is being used internally.

> I would like the submap/trajectory node/etc list to keep being published when the offline node enters its keep_waiting stage, when the simulation clock is standing still.

Can you elaborate what you mean by keep waiting?  Or do you mean `keep_running`?

---

Besides memory, one can argue if we want to adapt the publishing rate of topics to the simulation ("realistic" timing) or to a wall clock rate that's unrelated to the simulation time (convenient in some cases). This refers to the change of the meaning of the periods in simulation, as pointed out by Andre in his [comment](https://github.com/googlecartographer/cartographer_ros/pull/891#issuecomment-395099709) above.

ojura:
> This only accounts for batch processing. If you run with simulated time and e.g. pause the data publishing for a while, the memory shouldn't leak. This is not a user problem, also in the first case where finishing trajectories could make sense it's totally non-obvious why the memory increases unless you know that a wall timer is being used internally.

Okay, so it makes sense to use a regular timer for `PublishTrajectoryStates` (although you might miss seeing optimization results which happen while you're paused). But I really think we should have the rest of the topics on wall clock as before, so you can see optimization results on your submaps and trajectory markers even if the simulation time is standing still.

MichaelGrupp:
I get your point. Sounds like a good use case for a GlobalSlamResultCallback, is it? https://github.com/googlecartographer/cartographer/pull/1143

ojura:
There might be other examples when you might want to display the updated state. For example, the constraint builder might create new constraints, and this change should be visible (i.e. an updated constraint marker with the new constraints should be published) even before optimization is performed. This is why I would really prefer if we reverted back to wall clock for other publishers.

PR #898, opened 2018-06-13
Revert timers other than PublishTrajectoryStates back to being WallTimers

ojura:


ojura:
@gaschler @MichaelGrupp I am opening this PR as a suggestion and to continue discussion from #891.

To summarize, I believe we should keep timers other than the one for `PublishTrajectoryStates` as `WallTimers`, as before. This is because Cartographer's state can change even when the simulated time is standing still, and we want to visualize this. For example, optimization could get run and result in new trajectory and node poses, or the constraint builder may find new constraints. Additionally, we want to keep publishing everything when the offline node enters its `keep_running` stage.

MichaelGrupp:
Thanks for opening this PR, didn't have the time (or strong opinion) to come to a final conclusion regarding this. I'm totally fine with this revert if it removes headache for you. I'll let the reviewers decide ;)

One thing that I'd keep in mind is how we communicate / document the different publish periods in case of simulated time then.

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #901, opened 2018-06-15
Fix growing tf buffer

AlexReimann:
Decreases the time for writing `.ply` for me from _> 24h_ to _< 20min_ for a 8 GB bag.


SirVer:
I am lacking context, but that seems to make the tf_buffer entirely useless. Or am I missing something?

AlexReimann:
Yeah, nevermind I guess.
After looking into it a bit more, still not sure why this actually works at all for me...

Nevertheless there seems to be a problem with the tf_buffer growing too large.
Code wise it should adjust it's buffer automatically when a tf is looked up.
This seems to not happen for me for some reason..

AlexReimann:
Fixed by #879

PR #908, opened 2018-06-25
Improve internal naming of local SLAM data.

MichaelGrupp:
Prevents confusion with the new `TrajectoryState` and `GetTrajectoryStates()`
of the pose graph interface. The affected data is only local anyway.

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -223,7 +223,7 @@ cartographer_ros_msgs::SubmapList MapBuilderBridge::GetSubmapList() {
 
 std::unordered_map<int, MapBuilderBridge::LocalTrajectoryData>
 MapBuilderBridge::GetLocalTrajectoryData() {
-  std::unordered_map<int, LocalTrajectoryData> trajectory_states;
+  std::unordered_map<int, LocalTrajectoryData> local_trajectory_data;
   for (const auto& entry : sensor_bridges_) {
     const int trajectory_id = entry.first;
     const SensorBridge& sensor_bridge = *entry.second;
@@ -239,7 +239,7 @@ MapBuilderBridge::GetLocalTrajectoryData() {
 
     // Make sure there is a trajectory with 'trajectory_id'.
     CHECK_EQ(trajectory_options_.count(trajectory_id), 1);
-    trajectory_states[trajectory_id] = {
+    local_trajectory_data[trajectory_id] = {
         local_slam_data,
         map_builder_->pose_graph()->GetLocalToGlobalTransform(trajectory_id),
         sensor_bridge.tf_bridge().LookupToTracking(
@@ -247,7 +247,7 @@ MapBuilderBridge::GetLocalTrajectoryData() {
             trajectory_options_[trajectory_id].published_frame),
         trajectory_options_[trajectory_id]};
   }
-  return trajectory_states;
+  return local_trajectory_data;
 }
 
 visualization_msgs::MarkerArray MapBuilderBridge::GetTrajectoryNodeList() {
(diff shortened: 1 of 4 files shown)


cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -221,20 +221,20 @@ cartographer_ros_msgs::SubmapList MapBuilderBridge::GetSubmapList() {
   return submap_list;
 }
 
-std::unordered_map<int, MapBuilderBridge::TrajectoryState>
-MapBuilderBridge::GetTrajectoryStates() {
-  std::unordered_map<int, TrajectoryState> trajectory_states;
+std::unordered_map<int, MapBuilderBridge::LocalTrajectoryData>
+MapBuilderBridge::GetLocalTrajectoryData() {
+  std::unordered_map<int, LocalTrajectoryData> trajectory_states;
```
gaschler: rename to trajectory_data if you wish

MichaelGrupp: yes!

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -176,36 +176,36 @@ void Node::AddSensorSamplers(const int trajectory_id,
           options.landmarks_sampling_ratio));
 }
 
-void Node::PublishTrajectoryStates(const ::ros::TimerEvent& timer_event) {
+void Node::PublishLocalTrajectoryData(const ::ros::TimerEvent& timer_event) {
   carto::common::MutexLocker lock(&mutex_);
-  for (const auto& entry : map_builder_bridge_.GetTrajectoryStates()) {
-    const auto& trajectory_state = entry.second;
+  for (const auto& entry : map_builder_bridge_.GetLocalTrajectoryData()) {
+    const auto& trajectory_data = entry.second;
```
MichaelGrupp: I think it's fine to use trajectory_data here for brevity, it's clear from the context that it's local.

MichaelGrupp:
Thanks for the review. I also found some more places that could be renamed, they are also added now. See also my comment in the code.

cartographer_ros/cartographer_ros/node.h
```diff
@@ -152,7 +152,7 @@ class Node {
   void PublishSubmapList(const ::ros::WallTimerEvent& timer_event);
   void AddExtrapolator(int trajectory_id, const TrajectoryOptions& options);
   void AddSensorSamplers(int trajectory_id, const TrajectoryOptions& options);
-  void PublishTrajectoryStates(const ::ros::TimerEvent& timer_event);
+  void PublishLocalTrajectoryData(const ::ros::TimerEvent& timer_event);
```
gaschler: Reading this, I prefer "PublishTrajectoryData". It publishes the entire trajectory, the only thing local about it is it also looks at LocalSlamResult data to show that trajectory.

cartographer_ros/cartographer_ros/node.h
```diff
@@ -208,11 +208,11 @@ class Node {
...
-  // The timer for publishing trajectory states (i.e. pose transforms) is a
-  // regular timer which is not triggered when simulation time is standing
-  // still. This prevents overflowing the transform listener buffer by
-  // publishing the same transforms over and over again.
-  ::ros::Timer publish_trajectory_states_timer_;
+  // The timer for publishing local trajectory data (i.e. pose transforms and
+  // range data point clouds) is a regular timer which is not triggered when
+  // simulation time is standing still. This prevents overflowing the transform
+  // listener buffer by publishing the same transforms over and over again.
+  ::ros::Timer publish_local_trajectory_data_timer_;
```
gaschler: same here

cartographer_ros/cartographer_ros/node.h
MichaelGrupp: It publishes the scan matched pointcloud and the TF.

I think you mean `Node::PublishTrajectoryNodeList` which is not touched by this change?

gaschler: Sorry, you are right, I was mistaken about what PublishTrajectoryStates does.
Since there are no more open comments, should I merge?

MichaelGrupp: No problem, merging is fine for me.

PR #910, opened 2018-06-26
Use PoseGraphInterface::TrajectoryState from libcartographer

MichaelGrupp:
https://github.com/googlecartographer/rfcs/pull/35

- makes use of the trajectory state in `map_builder` and `node`
- adds a service to query the trajectory states
- follow-up to https://github.com/googlecartographer/cartographer/pull/1214
  that takes the deleted state into account in the `/finish_trajectory` service
  (could crash otherwise)

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -194,15 +194,9 @@ void MapBuilderBridge::HandleSubmapQuery(
   response.status.code = cartographer_ros_msgs::StatusCode::OK;
 }
 
-std::set<int> MapBuilderBridge::GetFrozenTrajectoryIds() {
-  std::set<int> frozen_trajectory_ids;
-  const auto node_poses = map_builder_->pose_graph()->GetTrajectoryNodePoses();
-  for (const int trajectory_id : node_poses.trajectory_ids()) {
-    if (map_builder_->pose_graph()->IsTrajectoryFrozen(trajectory_id)) {
-      frozen_trajectory_ids.insert(trajectory_id);
-    }
-  }
-  return frozen_trajectory_ids;
+std::map<int, ::cartographer::mapping::PoseGraphInterface::TrajectoryState>
+MapBuilderBridge::GetTrajectoryStates() {
+  return map_builder_->pose_graph()->GetTrajectoryStates();
 }
 
 cartographer_ros_msgs::SubmapList MapBuilderBridge::GetSubmapList() {
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -147,9 +147,12 @@ void MapBuilderBridge::FinishTrajectory(const int trajectory_id) {
   LOG(INFO) << "Finishing trajectory with ID '" << trajectory_id << "'...";
 
   // Make sure there is a trajectory with 'trajectory_id'.
-  CHECK_EQ(sensor_bridges_.count(trajectory_id), 1);
+  CHECK(GetTrajectoryStates().count(trajectory_id));
   map_builder_->FinishTrajectory(trajectory_id);
-  sensor_bridges_.erase(trajectory_id);
+  // Support cases with no subscribers, e.g. state visualization.
+  if (sensor_bridges_.count(trajectory_id)) {
+    sensor_bridges_.erase(trajectory_id);
+  }
 }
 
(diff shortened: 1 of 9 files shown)


MichaelGrupp:
As this is an (unplanned) early pull request for features of the yet unmerged RFC https://github.com/googlecartographer/rfcs/pull/35, let me know if you prefer to split it in separate PRs.

cartographer_ros/launch/visualize_pbstream.launch
```diff
@@ -23,7 +23,7 @@
           -configuration_basenames visualize_pbstream.lua
           -keep_running=true
           -load_state_filename $(arg pbstream_filename)
-          -load_frozen_state=false"
+          -load_frozen_state=true"
```
cschuet: this seems to not load the constraints that you might want to visualize.

MichaelGrupp: ahh, I already wondered. The reason I set it to true is this revised check that would fail otherwise: https://github.com/magazino/cartographer_ros/blob/use-trajectory-states-interface/cartographer_ros/cartographer_ros/node.cc#L633

It used to use the `is_active_trajectory_` map before which I removed. Will see if we can find a workaround...

MichaelGrupp:
@cschuet I added some graceful handling for the case with no subscribers (as in visualize_pbstream) in https://github.com/googlecartographer/cartographer_ros/pull/910/commits/a79fc684c1e3d8f8d996d51ea87462c47bfaa49e

Replacing the manually maintained `is_active_trajectory_` by TrajectoryState checks is worth it in my opinion.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -147,9 +147,12 @@ void MapBuilderBridge::FinishTrajectory(const int trajectory_id) {
   LOG(INFO) << "Finishing trajectory with ID '" << trajectory_id << "'...";
 
   // Make sure there is a trajectory with 'trajectory_id'.
-  CHECK_EQ(sensor_bridges_.count(trajectory_id), 1);
+  CHECK(GetTrajectoryStates().count(trajectory_id));
   map_builder_->FinishTrajectory(trajectory_id);
-  sensor_bridges_.erase(trajectory_id);
+  // Support cases with no subscribers, e.g. state visualization.
+  if (sensor_bridges_.count(trajectory_id)) {
+    sensor_bridges_.erase(trajectory_id);
```
cschuet: is erase() not a noop if the id does not exist?

cschuet:
@MichaelGrupp Thanks for the contribution.

cartographer_ros/cartographer_ros/map_builder_bridge.cc
MichaelGrupp: true. map::at() is the one that crashes

cartographer_ros/cartographer_ros/map_builder_bridge.h
```diff
@@ -77,7 +78,8 @@ class MapBuilderBridge {
       cartographer_ros_msgs::SubmapQuery::Request& request,
       cartographer_ros_msgs::SubmapQuery::Response& response);
 
-  std::set<int> GetFrozenTrajectoryIds();
+  std::map<int, ::cartographer::mapping::PoseGraphInterface::TrajectoryState>
```
gaschler: add /* trajectory_id */

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -550,10 +563,36 @@ int Node::AddOfflineTrajectory(
...
+          ::cartographer_ros_msgs::TrajectoryStates::ACTIVE);
+    } else if (entry.second == TrajectoryState::FINISHED) {
+      response.trajectory_states.trajectory_state.push_back(
+          ::cartographer_ros_msgs::TrajectoryStates::FINISHED);
+    } else if (entry.second == TrajectoryState::FROZEN) {
+      response.trajectory_states.trajectory_state.push_back(
+          ::cartographer_ros_msgs::TrajectoryStates::FROZEN);
+    } else {
+      response.trajectory_states.trajectory_state.push_back(
+          ::cartographer_ros_msgs::TrajectoryStates::DELETED);
```
gaschler: You could warn if there is an unknown state rather than returning DELETED.
Or use switch, which will fail to compile if not all states are covered.

cartographer_ros/cartographer_ros/map_builder_bridge.h
MichaelGrupp: thx

PR #914, opened 2018-06-29
Add internal metric families.

MichaelGrupp:
- minimal counter, gauge and histogram implementations
- metric family interfaces as defined in libcartographer
- serializable to ROS messages

Contributes to the corresponding RFC.

### DIFF
--- a/cartographer_ros/cartographer_ros/metrics/internal/counter.h
+++ b/cartographer_ros/cartographer_ros/metrics/internal/counter.h
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2018 The Cartographer Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CARTOGRAPHER_ROS_METRICS_INTERNAL_COUNTER_H
+#define CARTOGRAPHER_ROS_METRICS_INTERNAL_COUNTER_H
+
+#include "cartographer/metrics/counter.h"
+#include "cartographer_ros/metrics/internal/gauge.h"
+#include "cartographer_ros_msgs/Metric.h"
+
+namespace cartographer_ros {
+namespace metrics {
+
+class Counter : public ::cartographer::metrics::Counter {
+ public:
+  Counter() {}
+
+  Counter(const std::map<std::string, std::string>& labels) : gauge_(labels) {}
+
+  void Increment(const double by_value) override { gauge_.Increment(by_value); }
+
+  void Increment() override { gauge_.Increment(); }
+
+  double Value() const { return gauge_.Value(); }
+
+  cartographer_ros_msgs::Metric ToRosMessage() {
+    cartographer_ros_msgs::Metric msg = gauge_.ToRosMessage();
+    msg.type = cartographer_ros_msgs::Metric::TYPE_COUNTER;
+    return msg;
+  }
+
+ private:
+  Gauge gauge_;
+};
+
+}  // namespace metrics
+}  // namespace cartographer_ros
+
+#endif  // CARTOGRAPHER_ROS_METRICS_INTERNAL_COUNTER_H
--- a/cartographer_ros/cartographer_ros/metrics/internal/gauge.h
+++ b/cartographer_ros/cartographer_ros/metrics/internal/gauge.h
@@ -1,0 +1,77 @@
+/*
+ * Copyright 2018 The Cartographer Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef CARTOGRAPHER_ROS_METRICS_INTERNAL_GAUGE_H
+#define CARTOGRAPHER_ROS_METRICS_INTERNAL_GAUGE_H
+
+#include <atomic>
+#include <map>
+#include <string>
+
+#include "cartographer/metrics/gauge.h"
+#include "cartographer_ros_msgs/Metric.h"
+
+namespace cartographer_ros {
+namespace metrics {
+
+class Gauge : public ::cartographer::metrics::Gauge {
+ public:
+  Gauge() {}
+  Gauge(const std::map<std::string, std::string>& labels) : labels_(labels) {}
+
+  void Decrement(const double by_value) override { Add(-1. * by_value); }
+
+  void Decrement() override { Decrement(1.); }
+
+  void Increment(const double by_value) override { Add(by_value); }
+
+  void Increment() override { Increment(1.); }
+
+  void Set(double value) override {
+    double expected = value_.load();
+    while (!value_.compare_exchange_weak(expected, value)) {
+      ;
+    }
+  }
+  double Value() const { return value_.load(); }
+
+  cartographer_ros_msgs::Metric ToRosMessage() {
+    cartographer_ros_msgs::Metric msg;
+    msg.type = cartographer_ros_msgs::Metric::TYPE_GAUGE;
+    for (const auto& label : labels_) {
+      cartographer_ros_msgs::MetricLabel label_msg;
+      label_msg.key = label.first;
+      label_msg.value = label.second;
+      msg.labels.push_back(label_msg);
+    }
+    msg.value = Value();
+    return msg;
+  }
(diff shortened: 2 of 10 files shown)


cartographer_ros_msgs/msg/metrics/HistogramBucket.msg
```diff
@@ -0,0 +1,19 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# A histogram bucket between two boundaries i,j counts values x for which:
+#   bucket_boundary_i < x <= bucket_boundary_j
+# holds.
+float64 bucket_boundary
```
MichaelGrupp: @gaschler PTAL

cartographer_ros/cartographer_ros/metrics/internal/histogram.cc
```diff
@@ -0,0 +1,94 @@
...
+
+Histogram::Histogram(const BucketBoundaries& bucket_boundaries)
+    : Histogram({}, bucket_boundaries) {}
+
+void Histogram::Observe(double value) {
+  auto bucket_index =
+      std::distance(bucket_boundaries_.begin(),
+                    std::upper_bound(bucket_boundaries_.begin(),
+                                     bucket_boundaries_.end(), value));
+  // TODO: check lock contention and potential for atomic operations.
```
MichaelGrupp: @gaschler PTAL

cartographer_ros/cartographer_ros/metrics/internal/gauge.h
```diff
@@ -0,0 +1,77 @@
...
+  }
+
+  const std::map<std::string, std::string> labels_;
+  std::atomic<double> value_{0.};
+};
+
+}  // namespace metrics
+}  // namespace cartographer_ros
+
+#endif  // CARTOGRAPHER_ROS_METRICS_INTERNAL_GAUGE_H
```
MichaelGrupp: @gaschler Kept this one header-only because it's simple (also counter.h). The histogram is now split.

cartographer_ros_msgs/msg/metrics/HistogramBucket.msg
gaschler: Replace i, j by "this" boundary and "previous" or the like

cartographer_ros/cartographer_ros/metrics/internal/histogram.h
```diff
@@ -0,0 +1,62 @@
...
+constexpr double kInfiniteBoundary = std::numeric_limits<double>::infinity();
+
+using BucketBoundaries = ::cartographer::metrics::Histogram::BucketBoundaries;
+
+class Histogram : public ::cartographer::metrics::Histogram {
+ public:
+  Histogram(const std::map<std::string, std::string>& labels,
+            const BucketBoundaries& bucket_boundaries);
+
+  Histogram(const BucketBoundaries& bucket_boundaries);
```
gaschler: explicit

cartographer_ros/cartographer_ros/metrics/internal/histogram.h
```diff
@@ -0,0 +1,62 @@
...
+
+  void Observe(double value) override;
+
+  std::map<double, double> CountsByBucket();
```
gaschler: can these be private?

cartographer_ros/cartographer_ros/metrics/internal/histogram.cc
gaschler: Ok

cartographer_ros/cartographer_ros/metrics/internal/gauge.h
gaschler: Ok

cartographer_ros/cartographer_ros/metrics/internal/gauge.h
```diff
@@ -0,0 +1,77 @@
...
+
+#include "cartographer/metrics/gauge.h"
+#include "cartographer_ros_msgs/Metric.h"
+
+namespace cartographer_ros {
+namespace metrics {
+
+class Gauge : public ::cartographer::metrics::Gauge {
+ public:
+  Gauge() {}
```
gaschler: drop default constructor

cartographer_ros/cartographer_ros/metrics/internal/gauge.h
```diff
@@ -0,0 +1,77 @@
...
+  Gauge(const std::map<std::string, std::string>& labels) : labels_(labels) {}
```
gaschler: make explicit

cartographer_ros/cartographer_ros/metrics/internal/counter.h
```diff
@@ -0,0 +1,52 @@
...
+#include "cartographer/metrics/counter.h"
+#include "cartographer_ros/metrics/internal/gauge.h"
+#include "cartographer_ros_msgs/Metric.h"
+
+namespace cartographer_ros {
+namespace metrics {
+
+class Counter : public ::cartographer::metrics::Counter {
+ public:
+  Counter() {}
```
gaschler: drop default

cartographer_ros/cartographer_ros/metrics/internal/counter.h
```diff
@@ -0,0 +1,52 @@
...
+
+  Counter(const std::map<std::string, std::string>& labels) : gauge_(labels) {}
```
gaschler: make explicit

MichaelGrupp: ok :D

MichaelGrupp:
Changes added!

gaschler:
Is this ready to merge?

MichaelGrupp:
Yes, please merge

cartographer_ros/cartographer_ros/metrics/internal/gauge.h
```diff
@@ -0,0 +1,77 @@
...
+      label_msg.value = label.second;
+      msg.labels.push_back(label_msg);
+    }
+    msg.value = Value();
+    return msg;
+  }
+
+ private:
+  void Add(const double value) {
+    double current = Value();
```
CodeArno: @MichaelGrupp 
This is clearly a race condition: Two threads could read the current value, execute "current + value" in parallel, and then whoever calls Set first wins. The other increment is lost.

cartographer_ros/cartographer_ros/metrics/internal/gauge.h
```diff
@@ -0,0 +1,77 @@
...
+
+  void Decrement() override { Decrement(1.); }
+
+  void Increment(const double by_value) override { Add(by_value); }
+
+  void Increment() override { Increment(1.); }
+
+  void Set(double value) override {
+    double expected = value_.load();
+    while (!value_.compare_exchange_weak(expected, value)) {
```
CodeArno: @MichaelGrupp I'm not sure this is correct. What if two threads get the same "expected", and in one of them compare_exchange_weak succeeds, it returns true, and the while loop ends.
In the other, since the first thread already modified the value, compare_exchange_weak gets into an
infinite loop, until, by chance, someone sets value to the same value it was fetched in load.

Even if I'm misunderstanding, I would highly recommend not trying to be clever with an atomic in this implementation and just using a cartographer::common::Mutex to protect a plain old double value. That's much easier to understand and any potential performance problems would be negligible.

cschuet: @CodeArno I  think this code is correct: when compare_exchange_weak fails, the expected value is updated by a load.

https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange

cartographer_ros/cartographer_ros/metrics/internal/gauge.h
cschuet: Agreed, I think it would be better to

```
double expected = value_.load();
double new_value = expected + value;
 while (!value_.compare_exchange_weak(expected, new_value)) {
  new_value = expected + value;
}
```

MichaelGrupp: At a second glance at this code, I think we can even remove `Add` completely and directly use `std::atomic::fetch_add` and `std::atomic::fetch_sub` for increment and decrement, something like this:
```c++
  void Decrement(const double by_value) override { 
    value_.fetch_sub(by_value, std::memory_order_relaxed); 
  }

  void Decrement() override { Decrement(1.); }

  void Increment(const double by_value) override { 
    value_.fetch_add(by_value, std::memory_order_relaxed); 
  }

  void Increment() override { Increment(1.); }
```

MichaelGrupp: Do you agree?

cschuet: SG. I would rename by_value to e.g. decrement_value or something.

cartographer_ros/cartographer_ros/metrics/internal/gauge.h
MichaelGrupp: agree with @cschuet

cartographer_ros/cartographer_ros/metrics/internal/gauge.h
MichaelGrupp: Rollback: these operations are only supported for integral types....

wohe: Is there a reason you want to do this with atomics instead of using the less error-prone solution using a mutex? It seems unlikely you need any potential erformance benefit here.

MichaelGrupp: @CodeArno FYI metrics are now fixed with https://github.com/googlecartographer/cartographer_ros/commit/fb5b16d6d3b496482221bd60785be106419fa79d

PR #917, opened 2018-07-02
Register internal metrics and provide a public interface.

MichaelGrupp:
[RFC 24](https://github.com/googlecartographer/rfcs/blob/master/text/0024-monitoring-ros.md)

Public API:
  - adds `cartographer_ros::metrics::FamilyFactory`
  - compatible with `::cartographer::metrics::RegisterAllMetrics`

Public RPC interface:
  - adds the ROS service `/read_metrics`
  - response contains the latest values of all available metric families


MichaelGrupp:
Do you think the metrics should be optional? I.e. should we add an option `--collect_metrics=false` for the executables?

cschuet:
@MichaelGrupp Thanks, adding a command-line option with default ```false``` sounds good to me.

MichaelGrupp:
@cschuet Added the option. The service will return an `UNAIVALABLE` status code if it is not set.

MichaelGrupp:
Done.

PR #922, opened 2018-07-05
Simplify gauge and histogram implementation.

MichaelGrupp:
Use a mutex locker instead of atomic operations in gauge.

Remove unnecessary constructor overload from histogram.

### DIFF
--- a/cartographer_ros/cartographer_ros/metrics/internal/family.h
+++ b/cartographer_ros/cartographer_ros/metrics/internal/family.h
@@ -62,7 +62,7 @@ class HistogramFamily : public ::cartographer::metrics::Family<
  public:
   HistogramFamily(const std::string& name, const std::string& description,
                   const BucketBoundaries& boundaries)
-      : boundaries_(boundaries) {}
+      : name_(name), description_(description), boundaries_(boundaries) {}
 
   Histogram* Add(const std::map<std::string, std::string>& labels) override;
 
--- a/cartographer_ros/cartographer_ros/metrics/internal/gauge.h
+++ b/cartographer_ros/cartographer_ros/metrics/internal/gauge.h
@@ -32,20 +32,15 @@ class Gauge : public ::cartographer::metrics::Gauge {
   explicit Gauge(const std::map<std::string, std::string>& labels)
       : labels_(labels) {}
 
-  void Decrement(const double by_value) override { Add(-1. * by_value); }
+  void Decrement(const double value) override { Add(-1. * value); }
 
   void Decrement() override { Decrement(1.); }
 
-  void Increment(const double by_value) override { Add(by_value); }
+  void Increment(const double value) override { Add(value); }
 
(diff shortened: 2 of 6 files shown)


cartographer_ros/cartographer_ros/metrics/internal/gauge.h
```diff
@@ -63,8 +58,11 @@ class Gauge : public ::cartographer::metrics::Gauge {
 
  private:
   void Add(const double value) {
-    double current = Value();
-    Set(current + value);
+    double expected = value_.load();
+    double new_value = expected + value;
+    while (!value_.compare_exchange_weak(expected, new_value)) {
+      new_value = expected + value;
```
gaschler: you need to load expected again, otherwise the loop can be endless

cschuet: I think compare_exchange_weak does the load when it fails.

MichaelGrupp: expected is updated by reference

https://en.cppreference.com/w/cpp/atomic/atomic/compare_exchange
```
Atomically compares the object representation of *this with that of expected, and if those are bitwise-equal, replaces the former with desired (performs read-modify-write operation). Otherwise, loads the actual value stored in *this into expected (performs load operation). 
```

MichaelGrupp:
Since there was some discussion, do you think we should drop atomic completely and use a plain old mutex? @wohe @gaschler

wohe:
@MichaelGrupp +1 from my side.

MichaelGrupp:
Will change

cartographer_ros/cartographer_ros/metrics/internal/histogram.cc
```diff
@@ -44,7 +44,6 @@ void Histogram::Observe(double value) {
       std::distance(bucket_boundaries_.begin(),
                     std::upper_bound(bucket_boundaries_.begin(),
                                      bucket_boundaries_.end(), value));
-  // TODO: check lock contention and potential for atomic operations.
```
MichaelGrupp: @gaschler

cartographer_ros/cartographer_ros/metrics/internal/family.h
```diff
@@ -62,7 +62,7 @@ class HistogramFamily : public ::cartographer::metrics::Family<
  public:
   HistogramFamily(const std::string& name, const std::string& description,
                   const BucketBoundaries& boundaries)
-      : boundaries_(boundaries) {}
+      : name_(name), description_(description), boundaries_(boundaries) {}
```
MichaelGrupp: Just noticed that this was missing. Sneaked that into this PR if that's ok.

MichaelGrupp: Edit: removed responsible constructor (was a leftover from testing).

PR #926, opened 2018-07-06
Fix bug in FinishTrajectory logic

sebastianklose:
This PR adds additional bookkeeping for trajectories that we scheduled for
finishing.

In Node::RunFinalOptimization(...), we were calling FinishTrajectory for
every active trajectory (state == ACTIVE). Since the state only gets updated
once the corresponding worker for the FinishTrajectory task is
scheduled, we were potentially calling FinishTrajectory twice for a
single trajectory id.

Reproducible on master e.g. with 
```
roslaunch cartographer_ros offline_backpack_2d.launch bag_filenames:=b2-2016-02-02-14-01-56.bag no_rviz:=true
```

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -505,6 +505,7 @@ cartographer_ros_msgs::StatusResponse Node::FinishTrajectoryUnderLock(
     CHECK_EQ(subscribers_.erase(trajectory_id), 1);
   }
   map_builder_bridge_.FinishTrajectory(trajectory_id);
+  trajectories_scheduled_for_finish_.emplace(trajectory_id);
   const std::string message =
       "Finished trajectory " + std::to_string(trajectory_id) + ".";
   status_response.code = cartographer_ros_msgs::StatusCode::OK;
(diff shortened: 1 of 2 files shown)


sebastianklose:
@MichaelGrupp FYI - this was introduced after switching to use `TrajectoryState`

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -666,7 +668,8 @@ void Node::RunFinalOptimization() {
   {
     for (const auto& entry : map_builder_bridge_.GetTrajectoryStates()) {
       const int trajectory_id = entry.first;
-      if (entry.second == TrajectoryState::ACTIVE) {
+      if (entry.second == TrajectoryState::ACTIVE &&
+          trajectories_scheduled_for_finish_.count(entry.first) == 0) {
```
cschuet: move the check into FinishTrajectoryUnderLock() and add error message there talking about pending finish?

cschuet: And nice catch, BTW!

MichaelGrupp:
Oops, didn't notice that it behaves like that. Thanks.

cschuet:
@wally-the-cartographer merge

wally-the-cartographer:

PR #933, opened 2018-07-10
Allow zero pose_publish_period

gaschler:
To compare different SLAM software online, it is necessary to
disable tf broadcast.
Because we already have a parameter "pose_publish_period_sec",
we use a zero value here to turn off tf broadcast.


gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

Issue #929, opened 2018-07-06
Add warning if input topic is not present


gaschler:
Node constructs the input topic names from the lua config.

It is a very common mistake to not do proper remapping, in which case offline_node would just output

```
[ INFO] [1530886735.998898594]: I0706 16:18:55.000000 85316 submap_2d.cc:186] Added submap 1
[ INFO] [1530886735.999007268]: I0706 16:18:55.000000 85316 map_builder_bridge.cc:130] Added trajectory with ID '0'.
[ INFO] [1530886735.999070335]: I0706 16:18:55.000000 85316 offline_node.cc:274] Assigned trajectory 0 to bag filtered.bag
[ WARN] [1530886736.030330138, 1530736648.813927001]: W0706 16:18:56.000000 85316 ordered_multi_queue.cc:155] Queue waiting for data: (0, scan_1)
```

Even when one knows that remapping is necessary, it would be helpful if the node tells which input topics it expects, so a verbose warning would be better.

gaschler:
Fixed by #935

PR #935, opened 2018-07-11
Warn for possible topic mismatch

gaschler:
FIXES=#929

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -357,7 +357,9 @@ int Node::AddTrajectory(const TrajectoryOptions& options,
   AddExtrapolator(trajectory_id, options);
   AddSensorSamplers(trajectory_id, options);
   LaunchSubscribers(options, topics, trajectory_id);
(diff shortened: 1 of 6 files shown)


cartographer_ros/cartographer_ros/node.cc
```diff
@@ -357,6 +357,7 @@ int Node::AddTrajectory(const TrajectoryOptions& options,
   AddExtrapolator(trajectory_id, options);
   AddSensorSamplers(trajectory_id, options);
   LaunchSubscribers(options, topics, trajectory_id);
+  MaybeWarnAboutTopicMismatch();
```
kdaun: Checking for Mismatches here, leads to a racy behavior when starting a rosbag playbag at the same time. Resulting in info messages about missing topics that are actually there. Using a [ROS timer with the oneshot option](http://wiki.ros.org/roscpp/Overview/Timers) allows to check for the existence of the Topics e.g. 1s after the start of the trajectory, resolving the superfluous message.

gaschler:
Done, I delayed the check by 3 seconds and made it output warnings.

PR #938, opened 2018-07-12
Add .clang-format file.

wohe:
This adds a .clang-format file, so that git clang-format uses
Google style without the need to remember the commandline flag.
Similar to googlecartographer/cartographer#1249.


wohe:
@wally-the-cartographer merge

PR #939, opened 2018-07-13
Get rid of std::bind.

wohe:
`std::bind` is bug prone and should be avoided.
Lambdas are a more general and safer replacement.
Similar to googlecartographer/cartographer#1261.


wohe:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user wohe. Merge queue now has a length of 2.

PR #946, opened 2018-07-19
Add option to disable PoseExtrapolator

gaschler:
This is useful for tuning/debugging to rule out (simulated) time issues
(because published pose will then only depend on header times).
Another use case is when Cartographer runs on a separate machine
that has a different system clock than the sensors.


cartographer_ros/cartographer_ros/node_options.cc
```diff
@@ -40,6 +40,10 @@ NodeOptions CreateNodeOptions(
       lua_parameter_dictionary->GetDouble("pose_publish_period_sec");
   options.trajectory_publish_period_sec =
       lua_parameter_dictionary->GetDouble("trajectory_publish_period_sec");
+  if (lua_parameter_dictionary->HasKey("use_pose_extrapolator")) {
```
cschuet: Add use_pose_extrapolator = true; to lua files to make sure users are aware of this new feature?

gaschler:
@wally-the-cartographer merge

wally-the-cartographer:

PR #951, opened 2018-07-25
Follow the CMake update.

pifon2a:
https://github.com/googlecartographer/cartographer/pull/1332


cartographer_ros/CMakeLists.txt
```diff
@@ -48,7 +48,7 @@ include(FindPkgConfig)
 
 find_package(LuaGoogle REQUIRED)
 find_package(PCL REQUIRED COMPONENTS common)
-find_package(Eigen3 REQUIRED)
+find_package(Eigen REQUIRED)
```
gaschler: drop this change and rename find script

cartographer_ros/cmake/FindEigen.cmake
```diff
@@ -0,0 +1,30 @@
+# Copyright 2018 The Cartographer Authors
```
gaschler: Sure you need this file again?
I thought cartographer_ros/CMakeLists calls `include("${CARTOGRAPHER_CMAKE_DIR}/functions.cmake")` cartographer/cmake/functions.cmake which adds cartographer/cmake/modules

Same for cartographer_rviz

PR #955, opened 2018-07-26
Follow the Absl update.

pifon2a:


MichaelGrupp:
One question:

Abseil is not supported by Trusty/Indigo builds because of CMake, right? At least Trusty fails in the pipeline.

MichaelGrupp:
We're just using Indigo in the Travis pipeline of our cartographer_magazino playground repository, but would be good to know if we have to remove it or can fix it somehow. Wouldn't be tragic for to replace it though (with Stretch/Melodic) :)

gaschler:
> Abseil is not supported by Trusty/Indigo builds because of CMake, right?

We will depend on CMake 3.2 for building cartographer. In trusty, you can fulfill this by installing the Debian package `cmake3`

> At least Trusty fails in the pipeline.

We are working on this. In the current PR, abseil is not built automatically even though it is required transitively. Manually triggering `make standalone_absl && make cartographer_ros` works so it seems like a small dependency thing is missing in cartographer_ros/CMakeLists

MichaelGrupp:
Alright, then I think our Travis doesn't need a change if it's getting fixed here. It only includes the cartographer_ros repository, so we'll get your changes. But as I said, it's just for the playground, not affecting production code.

pifon2a:
please, use [CMAKE_STATIC_LIBRARY_SUFFIX](https://cmake.org/cmake/help/v3.0/variable/CMAKE_STATIC_LIBRARY_SUFFIX.html)

and

 [CMAKE_STATIC_LIBRARY_PREFIX(]https://cmake.org/cmake/help/v3.0/variable/CMAKE_STATIC_LIBRARY_PREFIX.html)

cschuet:
@pifon2a Holy moly! It passed!

PR #937, opened 2018-07-12
Follow googlecartographer/cartographer#1255

kdaun:

PR #981, opened 2018-08-12
Fix rosconsole-bridge CI issue

ojura:


ojura:
@cschuet  rosconsole-bridge being out of date was causing the build issue you observed in #937

cschuet:
@ojura Thanks for the PR. As discussed this will fail to compile without the point cloud changes. So I am closing this one.

PR #994, opened 2018-08-28
Removing unless from rviz_config argument

sebastianklose:
The roslaunch API throws an exception if setting the argument but not using it. 
This now forces any including launch files to provide an argument for it, even if it might not be used in the `no_rviz` case.


wally-the-cartographer:
Merge requested by authorized user kdaun. Merge queue now has a length of 1.

PR #1002, opened 2018-09-07
Refactor configuration parameters of publish periods.

MichaelGrupp:
Cleans up the names to be more descriptive:
- `submap_publish_period_sec` -> `submap_list_publish_period_sec`
- `pose_publish_period_sec` -> `transform_publish_period_sec`
- marker topics use `..._marker_publish_period_sec`

`visualization_msgs/MarkerArray` at a high frequency lead to high ROS traffic,
so it's useful to allow custom frequencies for _all_ marker topics if desired. Also:
- constraint marker publish period is not hardcoded anymore
- landmark markers now have their own publish period parameter


MichaelGrupp:
ToDo: follow-ups for cartographer_{toyota_hsr, fetch, magazino...}

MichaelGrupp:
Just an "FYI" for @ojura if you have anything to comment here, since you worked on this topic :)

cartographer_ros/configuration_files/revo_lds.lua
```diff
@@ -35,7 +35,9 @@ options = {
   lookup_transform_timeout_sec = 0.2,
   submap_publish_period_sec = 0.3,
```
ojura: While we're here, maybe we could clarify the names of these two? `submap_list_publish_period_sec` and `transform_publish_period_sec`?

MichaelGrupp: +1 for `submap_list_publish_period_sec`. But `lookup_transform_timeout_sec` seems fine to me, it's used as a timeout for in the TF bridge, not for publishing. What you mean is probably `pose_publish_period_sec`, which is used for the `publish_local_trajectory_data_timer_` that publishes the TF.

ojura: Yes, I was referring to `pose_publish_period_sec` -> `transform_publish_period_sec`.

PR #1021, opened 2018-09-15
Introduce base image for melodic to speed up CI

cschuet:


cschuet:
@MichaelGrupp @gaschler So this takes the melodic build time from > 50 min to 42'17. I was actually hoping that this would improve it even more. Somehow Travis wastes 6 minutes or so in downloading the base image. I haven't understood yet why the docker build cache is not helping.

Dockerfile.base.melodic
```diff
@@ -0,0 +1,43 @@
...
+# rosdep needs the updated package.xml files to install the correct debs.
+COPY cartographer_ros/package.xml catkin_ws/src/cartographer_ros/cartographer_ros/
+COPY cartographer_ros_msgs/package.xml catkin_ws/src/cartographer_ros/cartographer_ros_msgs/
+COPY cartographer_rviz/package.xml catkin_ws/src/cartographer_ros/cartographer_rviz/
+COPY scripts/install_debs.sh cartographer_ros/scripts/
+RUN cartographer_ros/scripts/install_debs.sh && rm -rf /var/lib/apt/lists/*
+
+# Install proto3.
+RUN /catkin_ws/src/cartographer/scripts/install_proto3.sh
+
```
MichaelGrupp: This doesn't include Ceres.

Dockerfile.melodic
```diff
@@ -41,9 +41,6 @@ COPY cartographer_rviz/package.xml catkin_ws/src/cartographer_ros/cartographer_r
 COPY scripts/install_debs.sh cartographer_ros/scripts/
 RUN cartographer_ros/scripts/install_debs.sh && rm -rf /var/lib/apt/lists/*
 
-# Install proto3.
-RUN /catkin_ws/src/cartographer/scripts/install_proto3.sh
-
```
MichaelGrupp: Remove also  `RUN cartographer_ros/scripts/install.sh --pkg ceres-solver` below and put it into the base build file?

PR #1040, opened 2018-10-02
Only use ROS log sink in pbstream_map_publisher_main.cc

MichaelGrupp:
Fixes double logging to stderr and rosout and makes it consistent with other
cartographer_ros nodes.

PR #1039, opened 2018-10-01
docs: Add documentation related to `bagfile_progress` publisher

alireza-hosseini:
Add documentations related to the feature introduced in #940.

PR #1041, opened 2018-10-02
Work around wstool bug in build with Docker base image.

MichaelGrupp:
We have already a cloned source space in the base image, which can be
buggy according to: https://github.com/vcstools/wstool/issues/77

This should unblock Travis and fix local builds.


MichaelGrupp:
#1039 and #1040 are blocked by this. They were opened after the `cartographer` repo has been changed the first time after the base image has been introduced in #1021. For whatever reason, `wstool` doesn't like that...

PR #1043, opened 2018-10-02
Restore offline node performance

ojura:
Revert "feat: Publish progress of processing the bagfile (#940)"
This reverts commit dd4d2af58b3a7f34b036abe5bf16b7d545b6e7d5.

Current performance of the offline node seems to be horribly broken after #940, as described in comment in #940. Please revert until fixed.


ojura:
@alireza-hosseini has quickly identified the problem, thanks!

PR #1044, opened 2018-10-02
Decrease asset writer progress log period

ojura:
With two 20 Hz lidars and a 200 Hz IMU, the progress reports were outputted every 500 seconds, which is really too sparse to be useful. Decrease this by a factor of 10.


ojura:
For me it was every 2 minutes of wall clock time, which is uncomfortably long :)

pifon2a:
@wally-the-cartographer merge

PR #1045, opened 2018-10-02
fix: Use an explicit message_counter instead of using `std::distance`

alireza-hosseini:
As @ojura reported and explained in #940, there is a critical issue with new changes introduced in the mentioned PR, which significantly slows down the offline node. 
- The reason of this problem was that In order to count the number of processed messages the `std::distance` function was used which is computationally expensive(O(n)).
- Instead, the former `log_counter_` and now `message_counter_` class variable which was used to print a message every X seconds is also employed to count the number of the processed (and also skipped) messages.


pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1066, opened 2018-10-17
Adding launch file arg for launch-prefix to offline nodes

sebastianklose:
Useful for debugging with gdb or profiling, e.g. with perf.


sebastianklose:
@wally-the-cartographer merge

ojura:
How would this work with gdb? Would you set the launch prefix to [`gdb -ex=r --args`](https://stackoverflow.com/a/18776339)?

sebastianklose:
@ojura yes -- also see [here](http://wiki.ros.org/roslaunch/Tutorials/Roslaunch%20Nodes%20in%20Valgrind%20or%20GDB) for more usage examples

PR #1089, opened 2018-11-05
Consider waiting trajectories with a sensor bridge as active.

MichaelGrupp:
Fixes https://github.com/googlecartographer/cartographer/issues/1367

Trajectories that didn't start SLAMing yet couldn't be finished, e.g. 
due to waiting for sensor data. They don't appear in the trajectory 
states list of the pose graph yet but already have a trajectory builder.


MichaelGrupp:
Yes, it looks like. I also noticed the failed CHECK, it's gone with this patch because the trajectories are finished now.

I'll mention the issue in the description here.

PR #1099, opened 2018-11-16
Make publishing tf optional, enable publishing PoseStamped

johuber:
If the output of cartographer should be used as an input to an additional sensor fusion method, using the published TF transform is inconvenient or in our specific use case even harmful because we don't want to add the raw cartographer output to our TF tree.

With this change it becomes optional to broadcast to /tf. Morever there is an option to publish the tracked frame pose as a PoseStamped message.
We add two new optional parameters:
- `publish_to_tf` if false no tf transform is broadcasted
-  `publish_tracked_pose` If set `true` a PoseStamped representing the position of the tracked pose w.r.t. map frame is published.

If default launchers and settings are used this PR causes no change in the behavior.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -40,6 +40,7 @@
 #include "cartographer_ros/time_conversion.h"
 #include "cartographer_ros_msgs/StatusCode.h"
 #include "cartographer_ros_msgs/StatusResponse.h"
+#include "geometry_msgs/PoseStamped.h"
 #include "glog/logging.h"
 #include "nav_msgs/Odometry.h"
 #include "ros/serialization.h"
@@ -113,6 +114,11 @@ Node::Node(
   constraint_list_publisher_ =
       node_handle_.advertise<::visualization_msgs::MarkerArray>(
           kConstraintListTopic, kLatestOnlyPublisherQueueSize);
+  if (node_options_.publish_tracked_pose) {
+    tracked_pose_publisher_ =
+        node_handle_.advertise<::geometry_msgs::PoseStamped>(
+            kTrackedPoseTopic, kLatestOnlyPublisherQueueSize);
+  }
   service_servers_.push_back(node_handle_.advertiseService(
       kSubmapQueryServiceName, &Node::HandleSubmapQuery, this));
   service_servers_.push_back(node_handle_.advertiseService(
@@ -271,32 +277,41 @@ void Node::PublishLocalTrajectoryData(const ::ros::TimerEvent& timer_event) {
         trajectory_data.local_to_map * tracking_to_local;
 
     if (trajectory_data.published_to_tracking != nullptr) {
-      if (trajectory_data.trajectory_options.provide_odom_frame) {
-        std::vector<geometry_msgs::TransformStamped> stamped_transforms;
+      if (node_options_.publish_to_tf) {
+        if (trajectory_data.trajectory_options.provide_odom_frame) {
+          std::vector<geometry_msgs::TransformStamped> stamped_transforms;
 
-        stamped_transform.header.frame_id = node_options_.map_frame;
-        stamped_transform.child_frame_id =
-            trajectory_data.trajectory_options.odom_frame;
-        stamped_transform.transform =
-            ToGeometryMsgTransform(trajectory_data.local_to_map);
-        stamped_transforms.push_back(stamped_transform);
+          stamped_transform.header.frame_id = node_options_.map_frame;
+          stamped_transform.child_frame_id =
+              trajectory_data.trajectory_options.odom_frame;
+          stamped_transform.transform =
+              ToGeometryMsgTransform(trajectory_data.local_to_map);
+          stamped_transforms.push_back(stamped_transform);
 
-        stamped_transform.header.frame_id =
-            trajectory_data.trajectory_options.odom_frame;
-        stamped_transform.child_frame_id =
-            trajectory_data.trajectory_options.published_frame;
-        stamped_transform.transform = ToGeometryMsgTransform(
-            tracking_to_local * (*trajectory_data.published_to_tracking));
-        stamped_transforms.push_back(stamped_transform);
+          stamped_transform.header.frame_id =
+              trajectory_data.trajectory_options.odom_frame;
+          stamped_transform.child_frame_id =
+              trajectory_data.trajectory_options.published_frame;
+          stamped_transform.transform = ToGeometryMsgTransform(
+              tracking_to_local * (*trajectory_data.published_to_tracking));
+          stamped_transforms.push_back(stamped_transform);
 
-        tf_broadcaster_.sendTransform(stamped_transforms);
-      } else {
-        stamped_transform.header.frame_id = node_options_.map_frame;
-        stamped_transform.child_frame_id =
-            trajectory_data.trajectory_options.published_frame;
-        stamped_transform.transform = ToGeometryMsgTransform(
-            tracking_to_map * (*trajectory_data.published_to_tracking));
-        tf_broadcaster_.sendTransform(stamped_transform);
+          tf_broadcaster_.sendTransform(stamped_transforms);
+        } else {
+          stamped_transform.header.frame_id = node_options_.map_frame;
+          stamped_transform.child_frame_id =
+              trajectory_data.trajectory_options.published_frame;
+          stamped_transform.transform = ToGeometryMsgTransform(
+              tracking_to_map * (*trajectory_data.published_to_tracking));
+          tf_broadcaster_.sendTransform(stamped_transform);
+        }
+      }
+      if (node_options_.publish_tracked_pose) {
+        ::geometry_msgs::PoseStamped pose_msg;
+        pose_msg.header.frame_id = node_options_.map_frame;
+        pose_msg.header.stamp = stamped_transform.header.stamp;
+        pose_msg.pose = ToGeometryMsgPose(tracking_to_map);
+        tracked_pose_publisher_.publish(pose_msg);
       }
     }
   }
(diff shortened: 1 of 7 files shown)


ojura:
@johuber I usually do `tf:=tf_null` when I don't want a node publishing stuff to TF. The only problem is when you want to subscribe to tf, but not publish - in that case introducing an another option is justified.

cartographer_ros/cartographer_ros/node_options.cc
```diff
@@ -24,6 +24,25 @@
 
 namespace cartographer_ros {
 
+std::array<double, 36> ReadCovariance(
+    ::cartographer::common::LuaParameterDictionary* const
+        lua_parameter_dictionary) {
+  std::array<double, 36> covariance = {};
```
MichaelGrupp: From a conceptual point of view, a pose covariance doesn't exist in Cartographer - a `PoseStamped` would be totally sufficient to describe the trajectory without TF. Adding constant parameters for the covariance to Cartographer's configuration therefore doesn't make sense, as these are specific to your type of filter pipeline and don't have any use in Cartographer.

But then, you can also just use a different TF topic name as @ojura suggested.

MichaelGrupp: > On the other hand I thought covariance is very related to the some cartographer settings, e.g. as voxel size.

Of course parameters like voxel size contribute to a systematic uncertainty. But defining a fixed pose covariance seems like a special requirement. And since it isn't used in Cartographer, it can be safely done outside of it. 

> So do you suggest to just use a different tf topic and obtaining the pose by subscribing to that topic?

I don't think this works since we also subscribe to TF (forget what I initially wrote). Maybe you can play around with the [`published_frame`](https://google-cartographer-ros.readthedocs.io/en/latest/configuration.html) parameter to publish to a child frame that doesn't affect the rest of the tree? (not quite sure at the moment if this will work). If not, I'd use a `PoseStamped`.

johuber: > Of course parameters like voxel size contribute to a systematic uncertainty. But defining a fixed pose covariance seems like a special requirement. And since it isn't used in Cartographer, it can be safely done outside of it.

Agreed, I'll remove the covariance part.

> I don't think this works since we also subscribe to TF (forget what I initially wrote). 

In my understanding tf is not actually used by cartographer, but static transforms (base_link->imu, base_link->lidar) are read directly from an URDF.
However, cartographer_ros makes use of the tf tree to transform from the `tracked_frame` to the `published_frame`. Hence, if we publish the `tracked_frame` it works fine. when using a different tf topic.

> Maybe you can play around with the `published_frame` parameter to publish to a child frame that doesn't affect the rest of the tree? (not quite sure at the moment if this will work). If not, I'd use a PoseStamped.

I tried that, but unfortunately it always leads to having either two separate trees (causes visualization issues and warnings) or having wrong information in the tree.

So my suggestion:
I will introduce a flag to publish a `PoseStamped` of the `tracked_frame`, but will remove the flag for disabling tf publishing, since this issue can be solved by remapping tf. (Although I still think it's unreasonable to crowd the TCP/IP stack with data we are not using..)

gaschler:
There is already a "hidden feature" to turn off publishing to tf by setting pose_publish_period_sec = 0.
Does this solve it?
I'm always hesitant to add new settings because we already have so many.

Also, what Michael explained, Cartographer does not use covariance. There were some experiments about this in the past but even in long hallways the covariance output did't show interesting principal components.

johuber:
@gaschler 

> There is already a "hidden feature" to turn off publishing to tf by setting pose_publish_period_sec = 0.
Does this solve it?
I'm always hesitant to add new settings because we already have so man

Does solve the problem witht the tf treet, but then I'd still need to introduce some form of pose publisher and introduce a new parameter that defines the rate of said publisher.
Also it would probably require renaming to something like `tf_publish_period_sec` and `pose_msg_publish_period_sec`, because the naming would be confusing otherwise.

johuber:
I changed it to publish a PoseStamped instead of PoseWithCovarianceStamped now

johuber:
is there still any interest in this change? Should I close the PR?

ojura:
Did you manage to solve your use case with the technique discussed above (tf remapping)?

johuber:
It is possible with the tf remapping solution, but I still believe it would be a nice feature if we also have traditional ROS-style message publishing. Because in our use case (and I believe in most others) we prefer to subscribe to a pose feedback in the form of a regular ROS-message.

If you're not interested in that you may close the PR.

ojura:
Disclaimer: I am not a maintainer, so I can't make that decision :) It may not be the case that they are disinterested, I think the maintainers are busy with other stuff since there's unfortunately been a lull with the repo activity in the past few months.

faizansid77:
Why is this still not merged?

johuber:
> Why is this still not merged?

was never approved.. I just rebased it to master and now appearently will need another approval.

rotu:
What’s the status of this PR? I could really use this functionality.

MichaelGrupp:
@bochen87 This PR was done at Enway, as far as I can tell. Do you want to rebase, sign it etc so we can do a fresh review? Since the PR was stale for a long time unfortunately and this seems to be a feature that would be welcomed by the community.

johuber:
> @bochen87 This PR was done at Enway, as far as I can tell. Do you want to rebase, sign it etc so we can do a fresh review? Since the PR was stale for a long time unfortunately and this seems to be a feature that would be welcomed by the community.
> 
> Also the PR message should be updated since we already decided against having a covariance.

Yes done at Enway by me. I think I will get to do this within this or next week, no new changes should be required

johuber:
@MichaelGrupp ready for review, did you offer to do this?

basically made no further changes (just rebased). It was already approved a while ago, but then the approval was lost due to rebasing

cartographer_ros/cartographer_ros/node_options.cc
```diff
@@ -40,6 +40,14 @@ NodeOptions CreateNodeOptions(
       lua_parameter_dictionary->GetDouble("pose_publish_period_sec");
   options.trajectory_publish_period_sec =
       lua_parameter_dictionary->GetDouble("trajectory_publish_period_sec");
+  if (lua_parameter_dictionary->HasKey("publish_to_tf")) {
+    options.publish_to_tf =
+        lua_parameter_dictionary->GetBool("publish_to_tf");
+  }
+  if (lua_parameter_dictionary->HasKey("publish_tracked_pose_msg")) {
+    options.publish_tracked_pose_msg =
+        lua_parameter_dictionary->GetBool("publish_tracked_pose_msg");
```
MichaelGrupp, line 49: I would remove `_msg` from the option name, it doesn't provide additional value.

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -113,6 +114,9 @@ Node::Node(
   constraint_list_publisher_ =
       node_handle_.advertise<::visualization_msgs::MarkerArray>(
           kConstraintListTopic, kLatestOnlyPublisherQueueSize);
+  tracked_pose_in_map_publisher_ =
+      node_handle_.advertise<::geometry_msgs::PoseStamped>(
+          kTrackedPoseTopic, kLatestOnlyPublisherQueueSize);
```
MichaelGrupp, line 119: Don't advertise if the feature is disabled by the node options.

cartographer_ros/cartographer_ros/node.h
```diff
@@ -188,6 +188,7 @@ class Node {
   ::ros::Publisher trajectory_node_list_publisher_;
   ::ros::Publisher landmark_poses_list_publisher_;
   ::ros::Publisher constraint_list_publisher_;
+  ::ros::Publisher tracked_pose_in_map_publisher_;
```
MichaelGrupp, line 191: One small thing: Since the topic is called `tracked_pose`, I would simplify the publisher name to `tracked_pose_publisher`.

MichaelGrupp (changes requested):
Besides my inline comments, please also add a description of this topic / parameter to the ROS API documentation, under the "published topics" section.
https://github.com/cartographer-project/cartographer_ros/blob/master/docs/source/ros_api.rst#published-topics

cartographer_ros/cartographer_ros/node_options.cc
johuber: done

cartographer_ros/cartographer_ros/node.cc
johuber: done

cartographer_ros/cartographer_ros/node.h
johuber: done

MichaelGrupp:
Please fix the build.

docs/source/configuration.rst
```diff
@@ -103,6 +103,12 @@ pose_publish_period_sec
   Interval in seconds at which to publish poses, e.g. 5e-3 for a frequency of
   200 Hz.
 
+publish_to_tf
+  Enable or disable providing of TF transforms.
+
+publish_tracked_pose_msg
+  Enable publishing of tracked pose as a `geometry_msgs/PoseStamped` to topic "tracked_pose".
```
MichaelGrupp, line 110: Double backticks are needed in RST, as you did in the other file.

MichaelGrupp: The other messages here in this file are links, so we should probably also add a link for PoseStamped.

johuber: I opted for
```
``geometry_msgs/PoseStamped`_
```
to be consistent with the rest of this file. But forgot the underscore before, sorry!

johuber: oh I only just realized what this link does. Will add that too!

MichaelGrupp: yes, RST is quite unituitive with simple inline links :smile: If in doubt, you can also build the html version locally using `sphinx-build docs/ out/` and check it.

Thank you

MichaelGrupp (approved):

PR #1157, opened 2019-01-14
Don't run final optimization in visualize_pbstream.launch

MichaelGrupp:
Replaces the offline node with the normal node.

The problem is that the offline node immediately runs a final
optimization with `visualize_pbstream.lua`, which is most likely not the
configuration that was used to generate the pbstream. This can lead to
effects like distortions in the map e.g. due to different weights, even
though the actual saved state is fine.

### DIFF
--- a/cartographer_ros/configuration_files/visualize_pbstream.lua
+++ b/cartographer_ros/configuration_files/visualize_pbstream.lua
@@ -18,4 +18,8 @@ POSE_GRAPH.constraint_builder.sampling_ratio = 0
 POSE_GRAPH.global_sampling_ratio = 0
 POSE_GRAPH.optimize_every_n_nodes = 0
 
+-- In case there are active publishers, drop all sensor data for visualization.
+options.imu_sampling_ratio = 0
+options.rangefinder_sampling_ratio = 0
+
 return options
--- a/cartographer_ros/configuration_files/visualize_pbstream.lua
+++ b/cartographer_ros/configuration_files/visualize_pbstream.lua
@@ -18,8 +18,4 @@ POSE_GRAPH.constraint_builder.sampling_ratio = 0
 POSE_GRAPH.global_sampling_ratio = 0
 POSE_GRAPH.optimize_every_n_nodes = 0
 
--- In case there are active publishers, drop all sensor data for visualization.
-options.imu_sampling_ratio = 0
-options.rangefinder_sampling_ratio = 0
-
 return options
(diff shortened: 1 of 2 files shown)


MichaelGrupp:
You only get a warning now that there are no input topics, but this can safely ignored.

Note: I think I already tried this and it didn't work a few months ago before https://github.com/googlecartographer/cartographer_ros/pull/1089 was merged, because back then the node couldn't finish trajectories that were just waiting for their topics. But now the node terminates cleanly ;)

ojura:
If you do not intend to optimize the loaded trajectory, why not freeze it?

MichaelGrupp:
> If you do not intend to optimize the loaded trajectory, why not freeze it?

Then the constraints wouldn't be displayed.

ojura:
Hmm, I see. I still feel loading an unfrozen trajectory for this purpose is not a robust solution, because the moment it gets optimized, it will get screwed up. I see two things which would both help solving this robustly.

The first would be to indeed load it frozen. [Not loading the constraints for frozen trajectories](https://github.com/googlecartographer/cartographer/blob/e816ce01b5c4fc9660a2b93852538e7f694896f3/cartographer/mapping/map_builder.cc#L375) is a memory saving-optimization, and perhaps deserves an option to disable it.

The second would be to able to ensure the same pose graph options are loaded. I actually implemented this in a private branch and it works well; if there's interest, we can look into upstreaming it. The idea is that the complete node options (including the map builder options) are stored in the .pbstream header, like in envisioned in the [rfc](https://github.com/googlecartographer/rfcs/blob/master/text/0015-serialization-header.md) from some time ago. You would then be able to tell Cartographer to use the node configuration from the .pbstream header instead of a new configuration_basename.

MichaelGrupp:
>  the moment it gets optimized, it will get screwed up

Yes, however: when would that happen? Unless there's sensor input it just does nothing. The only scenario I could think of is that you have a sensor publisher publishing the expected topics in parallel and the "visualization node" starts SLAMing, but this can be blocked by dropping all sensor data (see my last commit).

> Not loading the constraints for frozen trajectories is a memory saving-optimization, and perhaps deserves an option to disable it.

Unless there's another benefit I wouldn't want to change API only for this simple visualization tool...

> You would then be able to tell Cartographer to use the node configuration from the .pbstream header instead of a new configuration_basename.

This sounds useful, regardless of this PR.

ojura:
With the online node, instead of setting the sampler ratios to zero, I think it would make more sense just not to start the default trajectory. This is controlled by the flag `start_trajectory_with_default_topics`.

MichaelGrupp:
Thanks, I forgot about this option. Then it also doesn't warn about waiting for topics.

MichaelGrupp:
@gaschler merge?

ojura:
Related to my previous comment: https://github.com/googlecartographer/rfcs/pull/38

PR #1171, opened 2019-01-28
Fixed wrong Topic name for Landmarks

fprott:
The expected topic name for the cartographer seems to be landmark not landmarks even tough a list of landmarks is expected.


fprott:
I signed it!

doisyg:
The wrong topic name still appears on the read the doc site (it actually drove me crazy for an hour): https://google-cartographer-ros.readthedocs.io/en/latest/going_further.html

PR #1204, opened 2019-02-21
Add explicit --save_state_filename to offline node options.

MichaelGrupp:
This allows to serialize the state also when no bagfile is given, e.g.
when the offline node is used to run an optimization and/or trimming
configuration on a pbstream. Or simply when one wants to use a custom
name directly. This doesn't break compatibility with the previous CLI.

### DIFF
--- a/cartographer_ros/cartographer_ros/offline_node.cc
+++ b/cartographer_ros/cartographer_ros/offline_node.cc
@@ -22,6 +22,7 @@
 #include <sys/resource.h>
 #endif
 #include <time.h>
+
 #include <chrono>
 
 #include "absl/strings/str_split.h"
--- a/cartographer_ros/cartographer_ros/offline_node.cc
+++ b/cartographer_ros/cartographer_ros/offline_node.cc
@@ -62,6 +62,11 @@ DEFINE_string(load_state_filename, "",
               "a saved SLAM state.");
 DEFINE_bool(load_frozen_state, true,
             "Load the saved state as frozen (non-optimized) trajectories.");
+DEFINE_string(save_state_filename, "",
+              "Explicit name of the file to which the serialized state will be "
+              "written before shutdown. If left empty, the filename will be "
+              "inferred from the first bagfile's name as: "
+              "<bag_filenames[0]>.pbstream");
 DEFINE_bool(keep_running, false,
             "Keep running the offline node after all messages from the bag "
             "have been processed.");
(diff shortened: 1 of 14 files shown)


cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -269,9 +272,7 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
           ? playable_bag_multiplexer.PeekMessageTime()
           : ros::Time();
   while (playable_bag_multiplexer.IsMessageAvailable()) {
-    if (!::ros::ok()) {
-      return;
-    }
+    CHECK(::ros::ok());
```
MichaelGrupp: Changed these to checks, `!ros::ok()` should be considered as an error here IMO.

ojura: I put that there. This path is taken if you ctrl-C.

cartographer_ros/cartographer_ros/offline_node.cc
```diff
@@ -370,15 +371,19 @@ void RunOfflineNode(const MapBuilderFactory& map_builder_factory) {
   LOG(INFO) << "Peak memory usage: " << usage.ru_maxrss << " KiB";
 #endif
 
-  if (::ros::ok() && bag_filenames.size() > 0) {
-    const std::string output_filename = bag_filenames.front();
-    const std::string suffix = ".pbstream";
-    const std::string state_output_filename = output_filename + suffix;
+  // Serialize unless we have neither a bagfile nor an explicit state filename.
```
ojura: Same as above, please return the check for ros::ok(). We do not want to serialize if we received a sigabrt.

MichaelGrupp: Hmm... but this doesn't happen with `CHECK(ros::ok())`?

ojura: Isn‘t quitting with a failed CHECK due to ctrl-C a bit ugly than returning cleanly?

ojura: You have a point that it might be helpful to print an info message saying we are skipping serialization due to node shutting down. But a failed CHECK would still feel to ugly to me.

MichaelGrupp: For a batch processing tool like the offline node I consider any case where we run into "ROS not ok" before all the work is done as an error, so I see it as a bug if the offline node **doesn't** return with a non-zero exit code. For example: if the offline node is used in a pipeline and exits silently with "0" although it received a SIGINT in the middle of processing. 

An alternative without checks would be to log an error message and let the executable exit with `EXIT_FAILURE`.

ojura: Hmm, I see your point, you may be right about the return code. I have been mostly using the offline node as a regular ROS node - I spin it up, look at how it's doing, and maybe quit it if I am not happy. I reacted because I thought that barfing up a failed CHECK error message when you quit it using ctrl-C seemed odd to me - user's desire to quit is not an error per se.

Anyway, I just wanted to point that out, don't really know what really is the proper thing to do here. Maybe the maintainers will know :)

MichaelGrupp: > barfing up a failed CHECK error message

I agree that stacktrace and core dump is a bit overdramatic ;) I changed it back for now, it was just something I noticed while digging through the code and not super urgent.

In general the repeated checks for `ros::ok` are a bit odd. If for example final optimization takes very long it still blocks if you ctrl-c until the next `if (!ros::ok) { return; }` is hit.

pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user pifon2a. Merge queue now has a length of 2.

PR #1222, opened 2019-03-12
Service to query trajectory segments from the pose graph.

MichaelGrupp:
This `/trajectory_query` service allows to look up trajectory segments
from the pose graph.

This can be useful if one wants to get optimized trajectory data from
the current SLAM run and listening to live TF data would be too noisy.
For example, to stitch buffered point cloud data from a depth sensor
based on a recent localization trajectory segment of a robot.

Before, the pose graph trajectory nodes were not available except for
the trajectory marker topic (only positions, no orientation; inefficient)
or after serialization (which is not practical in live operation).

### DIFF
--- a/cartographer_ros/cartographer_ros/map_builder_bridge.cc
+++ b/cartographer_ros/cartographer_ros/map_builder_bridge.cc
@@ -265,17 +265,17 @@ void MapBuilderBridge::HandleTrajectoryQuery(
   if (!trajectory_states.count(request.trajectory_id)) {
     response.status.code = cartographer_ros_msgs::StatusCode::INVALID_ARGUMENT;
     response.status.message =
-        absl::StrCat("Requested trajectory with ID ",
-                     std::to_string(request.trajectory_id), " doesn't exist.");
+        absl::StrCat("Requested trajectory with ID ", request.trajectory_id,
+                     " doesn't exist.");
     return;
   } else if (trajectory_states.at(request.trajectory_id) ==
              cartographer::mapping::PoseGraphInterface::TrajectoryState::
                  DELETED) {
     response.status.code =
         cartographer_ros_msgs::StatusCode::RESOURCE_EXHAUSTED;
-    response.status.message = absl::StrCat(
-        "Requested trajectory with ID ", std::to_string(request.trajectory_id),
-        " has been deleted.");
+    response.status.message =
+        absl::StrCat("Requested trajectory with ID ", request.trajectory_id,
+                     " has been deleted.");
     return;
   }
   const cartographer::common::Time min_time = FromRos(request.min_time);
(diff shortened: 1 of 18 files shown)


cartographer_ros_msgs/srv/TrajectoryQuery.srv
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+int32 trajectory_id
+time min_time
+time max_time
+---
+cartographer_ros_msgs/StatusResponse status
+geometry_msgs/PoseStamped[] trajectory
```
MichaelGrupp: Using `geometry_msgs/PoseStamped` relative to the global map frame. In case of a deserialized trajectory, I think we can't recover the ROS tracking frame (no child frame ID)?

cartographer_ros_msgs/srv/TrajectoryQuery.srv
```diff
@@ -0,0 +1,20 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+int32 trajectory_id
+time min_time
+time max_time
```
MichaelGrupp: `min_time` and `max_time` make it more convenient for clients to get only segments they're interested in.
This could be also left out, and let the client filter it.

cartographer_ros_msgs/srv/TrajectoryQuery.srv
ojura: Remember my rgf about the API for storing user data in the proto stream? :-)

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -257,6 +258,62 @@ MapBuilderBridge::GetLocalTrajectoryData() {
...
+        absl::StrCat("Requested trajectory with ID ", request.trajectory_id,
+                     " doesn't exist.");
+    return;
+  } else if (trajectory_states.at(request.trajectory_id) ==
+             cartographer::mapping::PoseGraphInterface::TrajectoryState::
+                 DELETED) {
+    response.status.code =
+        cartographer_ros_msgs::StatusCode::RESOURCE_EXHAUSTED;
+    response.status.message =
+        absl::StrCat("Requested trajectory with ID ", request.trajectory_id,
```
MichaelGrupp: self-review: not much value of having an additional code path, can be simplified by combining with the other check to "Requested trajectory with ID x doesn't exist or has been deleted."

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -257,6 +258,62 @@ MapBuilderBridge::GetLocalTrajectoryData() {
...
+  const cartographer::common::Time max_time = FromRos(request.max_time);
+  const auto node_poses = map_builder_->pose_graph()->GetTrajectoryNodePoses();
+  for (const auto& node_id_data :
+       node_poses.trajectory(request.trajectory_id)) {
+    if (!node_id_data.data.constant_pose_data.has_value()) {
+      continue;
+    }
+    const cartographer::common::Time node_time =
+        node_id_data.data.constant_pose_data.value().time;
+    if (node_time < min_time || node_time > max_time) {
```
MichaelGrupp: self-review: remove min_time/max_time, this logic can be easily done on client side without losing functionality

cartographer_ros/cartographer_ros/map_builder_bridge.cc
```diff
@@ -258,6 +259,39 @@ MapBuilderBridge::GetLocalTrajectoryData() {
   return local_trajectory_data;
 }
 
+void MapBuilderBridge::HandleTrajectoryQuery(
+    cartographer_ros_msgs::TrajectoryQuery::Request& request,
+    cartographer_ros_msgs::TrajectoryQuery::Response& response) {
+  const auto trajectory_states = GetTrajectoryStates();
+  if (!trajectory_states.count(request.trajectory_id) ||
+      trajectory_states.at(request.trajectory_id) ==
```
pifon2a: please, use `find` and the resulting iterator instead of searching twice.

MichaelGrupp: I will use the new [`TrajectoryStateToStatus`](https://github.com/googlecartographer/cartographer_ros/commit/5fb4314c8b7f235af121c2bce2a5761e0adb8651#diff-e1f4eea55ec120f9e563c2b5d947449aR483) here and do the `find` optimization in `TrajectoryStateToStatus`.

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -476,50 +499,46 @@ bool Node::ValidateTopicNames(
...
+    status_response.message =
+        absl::StrCat("Trajectory ", trajectory_id, " doesn't exist.");
+    status_response.code = cartographer_ros_msgs::StatusCode::NOT_FOUND;
+    return status_response;
+  }
+
+  status_response.message =
+      absl::StrCat("Trajectory ", trajectory_id, " is in '",
+                   TrajectoryStateToString(it->second), "' state.");
+  if (valid_states.count(it->second)) {
```
pifon2a: nit: use `?:` instead of `if else` here.

pifon2a:
@wally-the-cartographer merge

wally-the-cartographer:

PR #1235, opened 2019-03-25
Fix segfault by changing the destruction order.

### LINKED ISSUE #1234: Segfault in RunWhenDoneCallback when --collect_metrics=true
I could reproduce a segfault by running an offline node with `--collect_metrics=true`. It seems to happen during shutdown after all computations are done (probably the reason why I never noticed it in all the roslaunch noise during shutdown).

Traceback obtained with `gdb` from core dump reveals that it happens in a worker thread in `RunWhenDoneCallback()`:
```
#0  0x0000000000000000 in ?? ()
#1  0x00000000005e5a62 in cartographer::mapping::constraints::ConstraintBuilder2D::RunWhenDoneCallback() ()
#2  0x00000000006c6b21 in cartographer::common::Task::Execute() ()
#3  0x00000000006c82f9 in cartographer::common::ThreadPool::DoWork() ()
#4  0x00007f30746f3c80 in ?? () from /usr/lib/x86_64-linux-gnu/libstdc++.so.6
#5  0x00007f307737c6ba in start_thread (arg=0x7f3055ffb700) at pthread_create.c:333
#6  0x00007f3073e5941d in clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:109
```

Because this doesn't happen when I set `--collect_metrics=false`, I suspect that the [gauge metric update](https://github.com/googlecartographer/cartographer/blob/8d5bf2a5dd823624c574359ba20558c9f0ab0323/cartographer/mapping/internal/constraints/constraint_builder_2d.cc#L286-L290) could be causing this. The error is gone if I comment that update out.

The root cause is probably that the family structure that owns the gauge metric (here, in `cartographer_ros`) is deleted too early and the worker thread still tries to access the dangling raw pointer to it.

MichaelGrupp:
The metrics registry is used as a raw pointer reference in map builder
and must outlive it.

Since C++ destroys in reverse declaration order, we have to put the
registry declaration before the map builder bridge.

Fixes #1234.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.h
+++ b/cartographer_ros/cartographer_ros/node.h
@@ -173,12 +173,13 @@ class Node {
   cartographer_ros_msgs::StatusResponse FinishTrajectoryUnderLock(
       int trajectory_id) EXCLUSIVE_LOCKS_REQUIRED(mutex_);
   void MaybeWarnAboutTopicMismatch(const ::ros::WallTimerEvent&);
-
+  
   const NodeOptions node_options_;
 
   tf2_ros::TransformBroadcaster tf_broadcaster_;
 
   absl::Mutex mutex_;
+  std::unique_ptr<cartographer_ros::metrics::FamilyFactory> metrics_registry_;
   MapBuilderBridge map_builder_bridge_ GUARDED_BY(mutex_);
 
   ::ros::NodeHandle node_handle_;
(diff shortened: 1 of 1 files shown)


sebastianklose:
@wally-the-cartographer merge

wally-the-cartographer:
Merge requested by authorized user sebastianklose. Merge queue now has a length of 1.

PR #1245, opened 2019-03-27
Simplify start_trajectory service (RFC-28)

mgladkova:
Details in https://github.com/googlecartographer/rfcs/blob/master/text/0028-simplify-ros-start-trajectory.md

PR #1262, opened 2019-04-12
Unify trajectory state checks for service handlers.

MichaelGrupp:
Some service handlers need to check if a trajectory is in a valid
state for the requested operation. If it's not, they return an error
response.

`Node::CheckTrajectoryState` removes code duplication in handlers.
It generates a human-readable response message and sets the
status code depending on whether the trajectory is in one of the 
`valid_states`.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -49,6 +49,12 @@
 
 namespace cartographer_ros {
 
+namespace carto = ::cartographer;
+
+using carto::transform::Rigid3d;
+using TrajectoryState =
+    ::cartographer::mapping::PoseGraphInterface::TrajectoryState;
+
 namespace {
 
 cartographer_ros_msgs::SensorTopics DefaultSensorTopics() {
@@ -80,13 +86,21 @@ template <typename MessageType>
           }));
 }
 
+std::string TrajectoryStateToString(const TrajectoryState trajectory_state) {
+  switch (trajectory_state) {
+    case TrajectoryState::ACTIVE:
+      return "ACTIVE";
+    case TrajectoryState::FINISHED:
+      return "FINISHED";
+    case TrajectoryState::FROZEN:
+      return "FROZEN";
+    case TrajectoryState::DELETED:
+      return "DELETED";
+  }
+  return "";
+}
+
 }  // namespace
-
-namespace carto = ::cartographer;
-
-using carto::transform::Rigid3d;
-using TrajectoryState =
-    ::cartographer::mapping::PoseGraphInterface::TrajectoryState;
 
 Node::Node(
     const NodeOptions& node_options,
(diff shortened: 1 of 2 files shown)


MichaelGrupp:
Currently only cleans up the `/finish_trajectory` code, but can be used in #1245 and potential future handlers.

cartographer_ros/cartographer_ros/node.h
```diff
@@ -174,6 +174,11 @@ class Node {
       int trajectory_id) EXCLUSIVE_LOCKS_REQUIRED(mutex_);
   void MaybeWarnAboutTopicMismatch(const ::ros::WallTimerEvent&);
 
+  // Helper function for service handlers that need to check trajectory states.
+  cartographer_ros_msgs::StatusResponse CheckTrajectoryState(
+      int trajectory_id,
+      std::set<cartographer::mapping::PoseGraphInterface::TrajectoryState>
```
gaschler: const &

cartographer_ros/cartographer_ros/node.h
```diff
@@ -174,6 +174,11 @@ class Node {
       int trajectory_id) EXCLUSIVE_LOCKS_REQUIRED(mutex_);
   void MaybeWarnAboutTopicMismatch(const ::ros::WallTimerEvent&);
 
+  // Helper function for service handlers that need to check trajectory states.
+  cartographer_ros_msgs::StatusResponse CheckTrajectoryState(
```
gaschler: naming: TrajectoryStateToStatus

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -466,51 +466,53 @@ bool Node::ValidateTopicNames(
...
-    return status_response;
-  } else if (trajectory_states.at(trajectory_id) == TrajectoryState::DELETED) {
-    const std::string error =
-        absl::StrCat("Trajectory ", trajectory_id, " has been deleted.");
-    LOG(ERROR) << error;
-    status_response.code =
-        cartographer_ros_msgs::StatusCode::RESOURCE_EXHAUSTED;
-    status_response.message = error;
-    return status_response;
+  cartographer_ros_msgs::StatusResponse check_response = CheckTrajectoryState(
```
gaschler: naming:response

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -466,51 +466,53 @@ bool Node::ValidateTopicNames(
...
+
+  if (!(trajectory_states.count(trajectory_id))) {
+    status_response.message =
+        absl::StrCat("Trajectory ", trajectory_id, " doesn't exist.");
+    status_response.code = cartographer_ros_msgs::StatusCode::NOT_FOUND;
+    return status_response;
+  }
+
+  const auto trajectory_state = trajectory_states.at(trajectory_id);
+  const std::map<TrajectoryState, std::string> state_to_string = {
```
gaschler: have helper function TrajectoryStateToString that uses switch without default so the compiler enforces the mapping to be complete

cartographer_ros/cartographer_ros/node.cc
MichaelGrupp: Changed it a bit differently: we can fill the already existing `status_response` variable

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -49,6 +49,12 @@
 
 namespace cartographer_ros {
 
+namespace carto = ::cartographer;
+
+using carto::transform::Rigid3d;
+using TrajectoryState =
+    ::cartographer::mapping::PoseGraphInterface::TrajectoryState;
```
MichaelGrupp: Moved this up to use the shortcut also in the anonymous namespace.

PR #1278, opened 2019-05-07
Remove unnecessary eigen_conversions dependency.

MichaelGrupp:
Seems to have been used in the past, but isn't needed anymore.


MichaelGrupp:
@googlebot pls check again (?)

Issue #913, opened 2018-06-29
Wrong pose assigned when start new trajectory with initial pose


iceicehyhy:
Hi,

I'm trying the pure localization mode now, and i'm able to end the previous trajectory and add a new trajectory with initial pose.

However, the initial pose being added does not follow the instruction I sent. It somehow assigns the robot to another point in the map.

iceicehyhy:
Hi,

thanks for your prompt reply.

the initial pose is set exactly on the map_frame that specified in the config file. What i've done is, I wrote a simple script subscribe to /initial_pose topic that RVIZ published, everytime i got the pose sent from RVIZ, i'll call the service to end the previous trajectory and start a new trajectory with the pose that i got from RVIZ.

The interesting thing is , the initial pose is working on some maps, but the pose is not correct for others. I wonder is there anything i need to do when I build the map? Or  any other thing i need to do before i call the /write_state service to save the serialized slam state?

MichaelGrupp:
Your approach sounds good, that's how I'm also doing it more or less. 

This shouldn't be affected by mapping. The only thing that could come to my mind is that the frame ID of the initial pose message doesn't match the frame ID of Cartographer's map frame.

---

Side note: since today you can use the `/get_trajectory_states` service to conveniently determine the active trajectories that need to be finished in your script. See https://github.com/googlecartographer/cartographer_ros/commit/8bbec6bff77a7e63669db34b9023ad1811e741fd ([msg](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros_msgs/msg/TrajectoryStates.msg) / [srv](https://github.com/googlecartographer/cartographer_ros/blob/master/cartographer_ros_msgs/srv/GetTrajectoryStates.srv))

iceicehyhy:
Coz the map I used is the grid map that generated by Cartographer occupancy grid node, is possible that the frame origin of the generated grid map differ from the origin of serialized state map ?

derekhe89:
I got the same problem, if you solve this problem, can you give me some advice?@ iceicehyhy

dan9thsense:
I would like to do the exact same thing.  @iceicehyhy can you share that script?

MichaelGrupp:
[...]

PR #1282, opened 2019-05-10
Update ROS API documentation with description of changed/new services

mgladkova:
Short description about new service /trajectory_query (#1222) and updated /start_trajectory (#1245).

### DIFF
--- a/docs/source/ros_api.rst
+++ b/docs/source/ros_api.rst
@@ -94,9 +94,11 @@ submap_query (`cartographer_ros_msgs/SubmapQuery`_)
   Fetches the requested submap.
 
 start_trajectory (`cartographer_ros_msgs/StartTrajectory`_)
-  Starts another trajectory by specifying its sensor topics and trajectory
-  options as an binary-encoded proto. Returns an assigned trajectory ID.
-  The ``start_trajectory`` executable provides a convenient wrapper to use this service.
+  Starts a trajectory using default sensor topics and the provided configuration. 
+  An initial pose can be optionally specified. Returns an assigned trajectory ID.
+
+trajectory_query (`cartographer_ros_msgs/TrajectoryQuery`_)
+  Returns the trajectory data from the pose graph.
 
 finish_trajectory (`cartographer_ros_msgs/FinishTrajectory`_)
   Finishes the given `trajectory_id`'s trajectory by running a final optimization.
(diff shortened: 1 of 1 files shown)


MarcelSoler:
Hello! 
I don't know If I'am asking in the proper place... But it's kind of related to the API and the documentation, so I'll ask it here.
I have a doubt regarding timestamp param of the initial_pose old service.

Testing with the old service I discovered that if I din't use {timestamp = 0 } the map coordinates didn't match my optimised trajectory https://github.com/googlecartographer/cartographer_ros/issues/913#issuecomment-483974223. Making the possibility of re-localising using the map coordinates impossible.

Reading https://github.com/googlecartographer/rfcs/blob/master/text/0028-simplify-ros-start-trajectory.md
I believe that you are taking about this parameter as keep time. And the conclusion of the discussion is that it won't be exposed in the ros service.

My question is how do I use the trajectory without optimisation (time 0 instead of current_time) without this param?  Is this taken into account in the refactor. If it is how can I use it properly?

Thank you!

MichaelGrupp:
@MarcelSoler The timestamp argument doesn't need to be set over the ROS service call anymore (to prevent confusion). It is now implicitly set to 0 in the service handler: https://github.com/googlecartographer/cartographer_ros/commit/92a8b81a8489dd71a73c6d838c560712311b1da5#diff-e1f4eea55ec120f9e563c2b5d947449aR571

So you don't need to worry about that anymore.

ojura:
Have you figured out what it does? :)

MarcelSoler:
@MichaelGrupp  Nice! I was worried that was non zero by default. Looking forward to try it!

MichaelGrupp:
@ojura The time is used to lookup an interpolated pose in the other trajectory, to which the relative pose is applied: see [here](https://github.com/googlecartographer/cartographer/blob/2adeb1f3beecf8427296ed2ff5a3e11925187521/cartographer/mapping/internal/2d/pose_graph_2d.cc#L1075). This means that:

* initial pose time == latest time -> relative pose is applied to end of other trajectory
* initial pose time == earliest time -> relative pose is applied to start of other trajectory (origin)
* somewhere in between -> relative pose is somewhere in between

Since the use cases for anything except the origin are probably rare (at least I don't come up with one), we just use the earliest time.

---

Moreover, the initial pose [updates the trajectory connectivity state](https://github.com/googlecartographer/cartographer/blob/2bd987ffb40e2d6667682c5c96432173f371094f/cartographer/mapping/internal/trajectory_connectivity_state.cc#L26) between the two trajectories in the pose graph - i.e. indicating that the two trajectories belong to the same disjoint set of connected trajectories. This speeds up constraint search.

cschuet:
Thanks @mgladkova !

Issue #1417, opened 2019-11-07
Run cartographer_ros without the map -> odom tf


davidbsp:
Is there a paremeter to run cartographer_ros without the map->odom tf?

For example, [Hector Mapping](http://wiki.ros.org/hector_mapping) has the following parameter:

`~pub_map_odom_transform (bool, default: true)`
`Determine if the map->odom transform should be published by the system.`

I would like to accomplish the same with cartographer_ros. 
Can this be done with a ROS param or must I change the code?

artemiialessandrini:
Also interested!

PR #1419, opened 2019-11-08
Added pub_map_odom_transform parameter (when provide_odom_frame is true)

davidbsp:
Solved issue https://github.com/googlecartographer/cartographer_ros/issues/1417.
Hope this can be merged to the main repo.

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -274,12 +274,15 @@ void Node::PublishLocalTrajectoryData(const ::ros::TimerEvent& timer_event) {
       if (trajectory_data.trajectory_options.provide_odom_frame) {
         std::vector<geometry_msgs::TransformStamped> stamped_transforms;
 
-        stamped_transform.header.frame_id = node_options_.map_frame;
-        stamped_transform.child_frame_id =
+        if (trajectory_data.trajectory_options.pub_map_odom_transform){
+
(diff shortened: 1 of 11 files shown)


davidbsp:
@googlebot I signed it!

MichaelGrupp:
I don't yet understand why one would need this, but maybe you can achieve what you want with the existing parameters already.

E.g. this config will directly publish map->base_link without publishing (or requiring) map->odom:

```lua
...

  map_frame = "map",
  tracking_frame = "base_link",
  published_frame = "base_link",
  odom_frame = "<unused>",
  provide_odom_frame = false,

...
```

Or, as a more odd config, if you then change `map_frame = odom` it should only publish odom->base_link. But IMO this only makes sense if Cartographer runs in a "pure odometry" mode with loop closure disabled.

davidbsp:
Hi @MichaelGrupp, thank you for the feedback.

The rationale behind this would be to use cartographer to provide pure local odometry, so we can use another approach at the higher level to provide global discrete localization. 

This would, for example, allow for an architecture that is robust to GPS dropouts (cartographer stops publishing poses if GPS dropouts occur).

The map->base_link solution wouldn't fit, as it would provide discrete pose jumps. 
However, your second suggestion makes sense (never occurred to me tbh). Will test it some day.

Cheers.

MichaelGrupp:
Closing this for now.

Issue #1334, opened 2019-06-25
Dockerfile for ros melodic is broken


razorheadfx:
Trying to build the current Dockerfile.melodic (at #92a8b81) like so:```docker build -t cartographer_broken -f Dockerfile.melodic .``` breaks.
```
Sending build context to Docker daemon  3.761MB
Step 1/25 : FROM eu.gcr.io/cartographer-141408/cartographer_ros_melodic_base
 ---> 4494cdd37eac
Step 2/25 : ARG CARTOGRAPHER_VERSION=master
 ---> Using cache
 ---> 12fba05dfdf3
Step 3/25 : ARG github_token
 ---> Using cache
 ---> 91a30d1d3d94
Step 4/25 : RUN apt-get update && apt-get install -y sudo && rm -rf /var/lib/apt/lists/*
 ---> Running in 86ce8532ce8f
Get:1 http://archive.ubuntu.com/ubuntu bionic InRelease [242 kB]
Get:2 http://archive.ubuntu.com/ubuntu bionic-updates InRelease [88.7 kB]
Get:3 http://security.ubuntu.com/ubuntu bionic-security InRelease [88.7 kB]
Get:4 http://archive.ubuntu.com/ubuntu bionic-backports InRelease [74.6 kB]
Get:5 http://packages.ros.org/ros/ubuntu bionic InRelease [4669 B]
Get:6 http://archive.ubuntu.com/ubuntu bionic/universe Sources [11.5 MB]
Err:5 http://packages.ros.org/ros/ubuntu bionic InRelease
  The following signatures couldn't be verified because the public key is not available: NO_PUBKEY F42ED6FBAB17C654
Get:7 http://archive.ubuntu.com/ubuntu bionic/main amd64 Packages [1344 kB]
Get:8 http://archive.ubuntu.com/ubuntu bionic/restricted amd64 Packages [13.5 kB]
Get:9 http://archive.ubuntu.com/ubuntu bionic/universe amd64 Packages [11.3 MB]
Get:10 http://security.ubuntu.com/ubuntu bionic-security/universe Sources [188 kB]
Get:11 http://archive.ubuntu.com/ubuntu bionic/multiverse amd64 Packages [186 kB]
Get:12 http://archive.ubuntu.com/ubuntu bionic-updates/universe Sources [325 kB]
Get:13 http://archive.ubuntu.com/ubuntu bionic-updates/universe amd64 Packages [1226 kB]
[...]

PR #1484, opened 2020-06-17
Fix broken build for melodic. Use official ros docker image.

bochen87:
Essentially revert #1021 as image seems to be fixed

Signed-off-by: Bo Chen <bo@enway.ai>


bochen87:
fixes #1379 #1334

MichaelGrupp:
Looks good. As [mentioned](https://github.com/cartographer-project/cartographer_ros/issues/1334#issuecomment-510806502)), the only reason we had introduced that base image was as a hotfix to reduce build times of Melodic that exceeded the limit of Travis.

As a follow-up we could consider whether some dependencies still need to be built from source all the time in CI or if sth. smarter could be done.

bochen87:
ok, so it passed the build in travis, so I'm going to merge it

PR #1485, opened 2020-06-25
Build Abseil dependency in CI.

wohe:
cartographer-project/cartographer#1711 removed
building Abseil from the Cartographer library.
We follow here the same approach as in
cartographer CI.

An alternative approach which is now possible is
adding an Abseil package for catkin and depending
on that.

### DIFF
--- a/Dockerfile.kinetic
+++ b/Dockerfile.kinetic
@@ -41,7 +41,8 @@ COPY cartographer_rviz/package.xml catkin_ws/src/cartographer_ros/cartographer_r
 COPY scripts/install_debs.sh cartographer_ros/scripts/
 RUN cartographer_ros/scripts/install_debs.sh && rm -rf /var/lib/apt/lists/*
 
-# Install proto3.
+# Install Abseil and proto3.
+RUN /catkin_ws/src/cartographer/scripts/install_abseil.sh
 RUN /catkin_ws/src/cartographer/scripts/install_proto3.sh
 
 # Build, install, and test all packages individually to allow caching. The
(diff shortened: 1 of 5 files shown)


pbaughman:
@wohe Can you clarify

> An alternative approach which is now possible is
> adding an Abseil package for catkin and depending
> on that.

Do you mean 
a) I, the end user can now expect abseil to be installed for me by a tool like rosdep because cartographer_ros lists it as a dependency

or

b) With some additional development effort in the future. it would be possible to add abseil as a catkin package and add a dependency to cartographer_ros's package.xml so that abseil is installed automatically by a tool like rosdep

Trying to understand the meaning of "now possible"  It reads as present tense, but I think it is meant as future-tense.

Thx - trying to understand what I need to do to fix my cartographer build that starting blowing up CI yesterday.  Right now I have it pinned to an old version which is sub-optimal.

wohe:
@pbaughman With now possible I mean that I tried it out: put Abseil into your catkin workspace, add a `package.xml`. In this case you do not need to build Abseil outside of catkin.

If you want an easy fix: Install `stow` and execute `cartographer/scripts/install_abseil.sh` to build and install Abseil on your system.

ojura:
@wohe should we add abseil to Cartographer's package.xml?

jonbinney:
+1 for adding an abseil dependency in cartographer's `package.xml` - it would certainly make it easier to build the code. The downside would be that if someone already had abseil manually installed (not built with its own `package.xml`) then catkin and rosdep wouldn't know how to satisfy the dependency.

On a related note, if this package gets released again, what is the preferred way to handle abseil dependency? Packaging abseil as its own ROS package is the "easiest" way because then that would build debs for abseil on the OSRF servers, for all relevant platforms.

PR #1488, opened 2020-07-01
Original ROS 2 Cartographer ROS port created from v1.0.0

clalancette:
This PR is a port of cartographer_ros to ROS 2. As you can see from the diff, it is actually quite a substantial change to cartographer_ros to make this work for ROS 2.

This code has been used in a few different robotics projects, so we know that at least the major parts of it work. There are some parts of it that could use a cleanup, as they workaround problems that may or may not still exist.

As-is, this PR doesn't apply to master. That's because this port was originally done against the 1.0.0 tag of this repository. Moving it forward would be possible, but would likely require some work.

I'm opening this PR as a starting point for the discussion on how to move this code into the upstream.  As of right now, the `dashing` branch here in https://github.com/cartographer-project/cartographer_ros is branched off of 1.0.0, so this PR applies without conflict.

There is also the open question on whether the cartographer project is willing to take over maintenance of this port and maintain/release it for ROS 2 distributions going forward.

Finally, I'll note that I don't have too much of time to devote to this. I'm doing this on discretionary time, so while I'm glad to do some simple work here, I won't have the time to do any major refactoring (like updating it to master).

Comments, questions, thoughts all welcome!


MichaelGrupp:
@clalancette Hi Chris, can you please reopen this PR and target the new branch `ros2-dashing-1.0.0`, as agreed in the meeting? Thanks.
I created this branch now and it's also based on the 1.0.0 tag. The "dashing" branch is removed now, which is also why this PR here got closed.

clalancette:
Sounds good, will go ahead and do that.

PR #1489, opened 2020-07-05
Add flag to round the occupancy grid values to the trinary interpretation

BarceloChristian:
- ROS navigation stack does not work correctly with the map currently published by Cartographer ROS.
- Autonomous exploration packages are not able to detect frontiers.
- MoveBase complains about TF needing to extrapolate into the future when running autonomous exploration nodes.
--------
- For the first two bullets, I added a FLAG `trinary_interpretation` into `occupancy_grid_node_main` and `pbstream_map_publisher_main` to round the values into `0` or `100` based on default `phit` and `pmiss`.
- For the third bullet, I added a FLAG to `node_main` and `offline_node` that allows setting a tolerance to the TF message. This is something similar to what AMCL and GMapping do.

Signed-off-by: Christian Barcelo <christianbarcelo@ekumenlabs.com>

### DIFF
--- a/cartographer_ros/cartographer_ros/msg_conversion.cc
+++ b/cartographer_ros/cartographer_ros/msg_conversion.cc
@@ -372,8 +372,8 @@ cartographer::transform::Rigid3d ComputeLocalFrameFromLatLong(
 
 std::unique_ptr<nav_msgs::OccupancyGrid> CreateOccupancyGridMsg(
     const cartographer::io::PaintSubmapSlicesResult& painted_slices,
-    const double resolution, const std::string& frame_id,
-    const ros::Time& time) {
+    const double resolution, const std::string& frame_id, const ros::Time& time,
+    bool trinary_interpretation) {
   auto occupancy_grid = absl::make_unique<nav_msgs::OccupancyGrid>();
 
   const int width = cairo_image_surface_get_width(painted_slices.surface.get());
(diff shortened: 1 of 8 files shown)


cartographer_ros/cartographer_ros/msg_conversion.cc
```diff
@@ -404,10 +404,20 @@ std::unique_ptr<nav_msgs::OccupancyGrid> CreateOccupancyGridMsg(
...
-      const int value =
-          observed == 0
-              ? -1
-              : ::cartographer::common::RoundToInt((1. - color / 255.) * 100.);
+
+      int value = -1;
+      if (observed != 0) {
+        const auto cell =
+            ::cartographer::common::RoundToInt((1. - color / 255.) * 100.);
+        if (trinary_interpretation) {
```
wohe, line 412: This is deviating from documentation of this message: http://docs.ros.org/api/nav_msgs/html/msg/OccupancyGrid.html

Did you consider improving the nav stack to handle this properly?

BarceloChristian: While it is correct that deviates from the documentation of OccupancyGrid, `map_server` as a standard in ROS uses that same message for publishing in this way.
It's true also that nav stack could be improved to properly work with the current approach that Cartographer has, but I want to bring an option that would allow ANY package developed to work with `map_server` (again, as a standard), to properly work with Cartographer too. As I mentioned this is the case of autonomous exploration packages as `explore-lite` and `frontier-exploration`. There are a lot of issues opened on the GitHub of those packages asking to work with Cartographer, and the same might be happening with others packages that  I don't know.

wohe: This sounds unconvincing. It seems you agree that the current implementation is correct. Also `map_server` which you mention seems to use the message as documented. http://wiki.ros.org/map_server even mentions `scale` and `raw` modes which would also break the other packages you are interested in it would seem at first glance. Is this not the case? I can only encourage you to investigate improvements in these packages. Not foremost to support the approach that Cartographer has, but to the support the OccupancyGrid message as it is documented.

PR #1493, opened 2020-07-11
Trajectory markers

jbarb365:
update for rover configuration


MichaelGrupp:
This is a custom assets writer and URDF config for some "rover". I think stuff like this should live in forks, unless it directly contributes to the official examples in the documentation (like the Google backpack data).

MichaelGrupp:
Closing for inactivity and given my previous argument.

PR #1494, opened 2020-07-15
Changes for ROS Noetic support

MichaelGrupp:
Adds a Dockerfile for ROS Noetic and fixes the package.xml

PR #1499, opened 2020-07-27
Prepare GMock support for Noetic.

wohe:
This follows Cartographer installing libgmock-dev.


wohe:
FYI @MichaelGrupp

wohe:
This is for #1494. It does:
1. Install `libgmock-dev` to use for GMock.
2. Saving `GMOCK_LIBRARIES` before it gets overwritten.
3. Explicitly add GMock and the GTest main to each test. Not strictly necessary right now, but good for a potential future where we do not have to rely on Cartographer's approach to find GTest and GMock anymore.

MichaelGrupp:
Thanks

PR #1510, opened 2020-08-28
Remove legacy Jenkins quality pipeline.

MichaelGrupp:
- it has been used only at Google for a quality pipeline
- the worker.py is Python 2 and closely tied to Google Cloud, so I don't
  believe there is much to be reused by someone else nowadays and not
  worth the effort to port to Python 3.


MichaelGrupp:
I have access to the dashboard and it is inactive or broken since a while. So we can safely remove this.

image

bochen87 (approved):
makes sense, let's remove!

PR #1525, opened 2020-10-13
Follow cartographer-project/cartographer#1759.

wohe:


wohe:
@bochen87 @MichaelGrupp This is needed for cartographer-project/cartographer#1759 but it also works for the existing Cartographer code where it is a no-op. So we can merge it already in advance.

MichaelGrupp (approved):
lgtm, please rebase

PR #1530, opened 2020-10-22
Default to master in rosinstall (fixes #1122).

### LINKED ISSUE #1122: rosinstall file defaults to 1.0.0
When trying to run the pure localization demo 2D from [here](https://google-cartographer-ros.readthedocs.io/en/latest/demos.html#pure-localization) I get `Key 'pure_localization_trimmer' was used the wrong number of times.` (I first encountered this issue when trying to run pure localization on [my own code](https://github.com/Roboy/cartographer_ros/blob/roboy/cartographer_ros/configuration_files/roboy_localization.lua), where [your updated syntax](https://google-cartographer-ros.readthedocs.io/en/latest/going_further.html#localization-only) on pure localization didn't seem to work.)

Are you able to reproduce the behavior of the demo (let's not talk about my code for now) mentioned above? 

Best,
Chris

MichaelGrupp:
This makes much more sense than 1.0.0, which is counter-intuitive.

If someone is building from source they usually do it to build the
latest bleeding-edge version.


johuber (approved):
It does make sense to me too!

PR #1518, opened 2020-09-24
Follow cartographer-project/cartographer#1750.

wohe:
Signed-off-by: Wolfgang Hess <whess@lyft.com>

PR #1532, opened 2020-10-23
Remove ceres-solver from cartographer_ros.rosinstall.

wohe:
Recent versions no longer build ceres-solver and instead
depend on libceres-dev. See also #1518.


MichaelGrupp (approved):
Thanks, I noticed this also yesterday as written in the linked ticket.

PR #1540, opened 2020-10-28
Add load / delete ROS services.

MichaelGrupp:


MichaelGrupp:
Requires https://github.com/cartographer-project/cartographer/pull/1767 for deleting frozen trajectories.

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -725,6 +729,39 @@ bool Node::HandleReadMetrics(
...
+      "Scheduled trajectory ", request.trajectory_id, " for deletion.");
+  LOG(INFO) << response.status.message;
+  return true;
+}
+
+bool Node::HandleLoadStateFromFile(
+    cartographer_ros_msgs::LoadStateFromFile::Request& request,
+    cartographer_ros_msgs::LoadStateFromFile::Response& response) {
+  absl::MutexLock lock(&mutex_);
+  // TODO(MichaelGrupp): this crashes if the file is invalid.
```
johuber, line 755: why not handle the case and return it with the status?

Should be easy with `try / catch` rigth?

johuber:
The load state from file worked as expected.

I got it to crash with the delete call though, following error:
This happened consistently about 15s after the call to the delete service (of a `FINISHED` trajectory). 
```
F0128 14:04:08.248204 32214 problem_impl.cc:635] Parameter block not found: 0x7f67e0035b80. You must add the parameter block to the problem before you can set a lower bound on one of its components.
[FATAL] [1611839048.248835927][/cartographer/cartographer_node][/home/jonathan/cartographer_ws/src/cartographer_ros/cartographer_ros/cartographer_ros/ros_log_sink.cc:63]: F0128 14:04:08.000000 32214 problem_impl.cc:635] Parameter block not found: 0x7f67e0035b80. You must add the parameter block to the problem before you can set a lower bound on one of its components.
```

When I called delete onto a frozen trajectory nothing seemed to happen. Is that expected because there are still active constraints between the trajectories? 
Side node; I tested this in pure localization mode, while the robot was not localized with respect to the frozen trajectory

MichaelGrupp:
Will address your points at some later point and let you know, I currently don't have too much time for this.

Wang-Bo-xuan:
When deleting the same submaps of multiple trajectories, because the reference number is not checked in the pose graph, repeated delete will cause the program to crash.

Issue #1536, opened 2020-10-26
New ROS 2 port of cartographer_ros - tracking issue


MichaelGrupp:
We plan to make a new ROS 2 port of cartographer_ros that is in sync with master.

There is a port of an older version that can be used as an inspiration for a new port: https://github.com/cartographer-project/cartographer_ros/pull/1488

This issue aims to track progress and provide an entry point for people who are interested to contribute.

(_work in progress_)

3473f:
Would definitely love to get involved as I have been getting my hands dirty with the source code recently. Any leads on where to start would be appreciated.

Serafadam:
Hi, I'd also like to help, for starters, I am currently using Cartographer in one of my projects, from the ros2 repository. If I want to move to ROS2 Foxy, I'll also need to upgrade rplidar repository to this distro. As for creating this version, will there be some kind of release? I'd rather not start working before some huge feature gets added. As for the way of working, I think the best would be to branch off the latest commit here, and base work on https://github.com/cartographer-project/cartographer_ros/pull/1488.

How does that sound?

MichaelGrupp:
Hi @3473f and @Serafadam, thanks for joining the discussion! I have just updated the description with more details that we discussed this week in a maintainers meeting. Would be great if you can have a look and let us know what you think and if this is something you would like to contribute to.

---

> As for creating this version, will there be some kind of release?

Yes, that's the goal. To have a new release and then in the future also more frequent releases.

> As for the way of working, I think the best would be to branch off the latest commit here, and base work on #1488.

I also think so, as I wrote now in the description.

3473f:
Hi @MichaelGrupp, sorry for the late reply! that definitely sounds interesting. I have a lot of my plate in the next two weeks, after which I can start contributing regularly to this port.

> Create a new branch from master and apply similar changes like the ones that were added in the original ROS2 port.
I'd rather follow this approach as well. The only case I'd find starting from the current ROS2 port useful is if it was a few commits behind, but it seems that Cartographer has changed a lot since then.

airballking:
@MichaelGrupp Great initiative! My team has good experience using Cartographer and wants to switch our current robot system from ROS1 to ROS2. Right now, Cartographer is the only dependency that is still blocking this migration. So, we are willing to help. Let us know how we can contribute!

MichaelGrupp:
Hi, to give you an update: I actually started doing some things myself already:

[...]

PR #1541, opened 2020-10-30
Original ROS 2 Cartographer ROS port created from v1.0.0

clalancette:
This PR is a port of cartographer_ros to ROS 2. As you can see from the diff, it is actually quite a substantial change to cartographer_ros to make this work for ROS 2.

This code has been used in a few different robotics projects, so we know that at least the major parts of it work. There are some parts of it that could use a cleanup, as they workaround problems that may or may not still exist.

As-is, this PR doesn't apply to master. That's because this port was originally done against the 1.0.0 tag of this repository. Moving it forward would be possible, but would likely require some work.

I'm opening this PR as a starting point for the discussion on how to move this code into the upstream. As of right now, the ros2-dashing-1.0.0 branch here in https://github.com/cartographer-project/cartographer_ros is branched off of 1.0.0, so this PR applies without conflict.

Comments, questions, thoughts all welcome!


MichaelGrupp:
@wohe @bochen87 Since this will be an archived branch (see #1536) I want to preserve the commit history. But I can't do a simple fast-forward merge, only squash is possible. Maybe you have more rights? It would only be needed once for this archive branch.

PR #1542, opened 2020-11-02
Port cartographer_ros_msgs to ROS2.

MichaelGrupp:
Includes changes made in the original 1.0 version and adaptation to
new messages and services that were added since then.

Original authors of 1.0 port:

Chris Lalancette (Open Robotics)
Mikael Arguedas (Open Robotics)
Shane Loretz (Open Robotics)
Darby Lim (ROBOTIS)

Signed-off-by: Michael Grupp <grupp@magazino.eu>

PR #1543, opened 2020-11-02
Set up an initial CI with colcon on ROS2 Dashing & Foxy.

MichaelGrupp:


MichaelGrupp:
Build and test looks fine, see the Travis logs. I would like to merge this before doing further steps like #1542. 

@clalancette Maybe you can also have a look. I literally started working ROS2 this week out of interest ;)

scripts/prepare_dev_workspace.sh
```diff
@@ -19,9 +19,9 @@ set -o verbose
 
 . /opt/ros/${ROS_DISTRO}/setup.sh
 
-# Create a new workspace in 'catkin_ws'.
-mkdir -p catkin_ws/src
-cd catkin_ws/src
+# Create a new workspace in 'dev_ws'.
+mkdir -p dev_ws/src
+cd dev_ws/src
 wstool init
```
clalancette, line 25: This is probably a good opportunity to switch from `wstool` -> `vcstool`, since `wstool` is indeed deprecated (and now unmaintained).

Dockerfile.dashing
```diff
@@ -0,0 +1,66 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+FROM osrf/ros:dashing-desktop
+
+ARG CARTOGRAPHER_VERSION=master
```
clalancette, line 17: I'm not really familiar with how cartographer CI generally works, but shouldn't this be `ros2-dashing`?  Same for Dockerfile.foxy below.

clalancette (commented):
I pointed out a couple of minor things I noticed here.

scripts/prepare_dev_workspace.sh
MichaelGrupp: I would do this this as a separate change because it also needs to be done in the ROS1 code. I made a todo issue: #1549

Dockerfile.dashing
MichaelGrupp: This just means that we build against `master` of the cartographer core library.

PR #1556, opened 2020-11-10
Check if we already published a transformation at the same timestamp.

### LINKED ISSUE #1555: geometry2/tf2 warning TF_REPEATED_DATA
Git commit hash: `2712067e0563229248ac60c3e7be9b55c462bdce` (and all previous)
 
This issue is linked with the following issue: https://github.com/ros/geometry2/issues/467

A change in geometry2 package was introduced where there is a warning when multiple transforms between the same frames are published with the same timestamp.

This happens in simulation, but I did not test it on the robot yet, but from looking at the source code, I think the problem is specifically with the simulation at the following lines:
https://github.com/cartographer-project/cartographer_ros/blob/2712067e0563229248ac60c3e7be9b55c462bdce/cartographer_ros/cartographer_ros/node.cc#L258-L263

My guess: if using the pose_extrapolator, simulated time can(?) trigger the timer event multiple times if `trajectory_publish_period_sec` is sufficiently small which results in repeated TF publishing (because in the first iteration extrapolator advances which results in `stamped_transform.header.stamp` being the same in the next iteration)

The resulting repeated transform is ignored by the tf2 buffers (new standard behavior in ROS noetic), so the only consequence is a lot of generated warnings (which makes the roslog unreadable)

stribor14:
fix #1555

Signed-off-by: stribor14 <mirko.kokot@romb-technologies.hr>

### DIFF
--- a/cartographer_ros/cartographer_ros/node.cc
+++ b/cartographer_ros/cartographer_ros/node.cc
@@ -261,6 +261,15 @@ void Node::PublishLocalTrajectoryData(const ::ros::TimerEvent& timer_event) {
         node_options_.use_pose_extrapolator
             ? ToRos(now)
             : ToRos(trajectory_data.local_slam_data->time);
+
+    // Suppress publishing if we already published a transform at this time.
+    // Due to 2020-07 changes to geometry2, tf buffer will issue warnings for
+    // repeated transforms with the same timestamp.
+    if (last_published_tf_stamps_.count(entry.first) &&
+        last_published_tf_stamps_[entry.first] == stamped_transform.header.stamp)
+      continue;
+    last_published_tf_stamps_[entry.first] = stamped_transform.header.stamp;
+
     const Rigid3d tracking_to_local_3d =
         node_options_.use_pose_extrapolator
             ? extrapolator.ExtrapolatePose(now)
--- a/cartographer_ros/cartographer_ros/node.h
+++ b/cartographer_ros/cartographer_ros/node.h
@@ -214,6 +214,7 @@ class Node {
 
   // These are keyed with 'trajectory_id'.
   std::map<int, ::cartographer::mapping::PoseExtrapolator> extrapolators_;
+  std::map<int, ::ros::Time> last_published_tf_stamps_;
   std::unordered_map<int, TrajectorySensorSamplers> sensor_samplers_;
   std::unordered_map<int, std::vector<Subscriber>> subscribers_;
   std::unordered_set<std::string> subscribed_topics_;


cartographer_ros/cartographer_ros/node.cc
```diff
@@ -261,6 +261,16 @@ void Node::PublishLocalTrajectoryData(const ::ros::TimerEvent& timer_event) {
         node_options_.use_pose_extrapolator
             ? ToRos(now)
             : ToRos(trajectory_data.local_slam_data->time);
```
MichaelGrupp, line 263: I think it would be better to change this here and just use
```c++
stamped_transform.header.stamp = ToRos(now)
```
always.

I experienced similar timing issues when using Cartographer with `use_pose_extrapolator == false` in a Gazebo simulator and solved it like this (patch is currently only in a private fork of this repo).

MichaelGrupp (changes requested):

cartographer_ros/cartographer_ros/node.cc
stribor14: I just tried this fix and it doesn't work for me, warning is happening again

ojura: Hi Michael, I think that the data should be accurately stamped - if you're using the external pose extrapolator, the published extrapolated pose data should have a newer stamp; otherwise, it should be stamped as it is, the last lidar pose coming from local slam. And with a high enough publishing frequency, you can run into the same stamp if you have a `use_sim_time` clock.

The solution is that wherever you're using the external pose extrapolator or not, the stamp of the published poses should be enforced to be scrictly increasing. I think this PR accurately represents this intent.

@stribor14 you're missing a dot at the second row of the comment.

MichaelGrupp: @ojura Thanks for the clarification. I would be fine with merging this, I will only add a few comments for smaller nits.

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -261,6 +261,16 @@ void Node::PublishLocalTrajectoryData(const ::ros::TimerEvent& timer_event) {
...
+
+    // Suppress publishing if we already published a transform at this time.
+    // Due to 2020-07 changes to geometry2, tf buffer will issue warnings for
+    // repeated transforms with the same timestamp
+    if (last_published_tf_stamp_.find(entry.first) !=
+            last_published_tf_stamp_.end() &&
```
MichaelGrupp, line 269: If you want to check if a simple `std::map` has the key, you can simplify this to 
```
if (last_published_tf_stamps.count(entry.first) && ...
```
(`last_published_tf_stamps.count(entry.first) != 0` would also be possible, but the count gets casted to false if it's zero.)

cartographer_ros/cartographer_ros/node.cc
```diff
@@ -261,6 +261,16 @@ void Node::PublishLocalTrajectoryData(const ::ros::TimerEvent& timer_event) {
         node_options_.use_pose_extrapolator
             ? ToRos(now)
             : ToRos(trajectory_data.local_slam_data->time);
+
+    // Suppress publishing if we already published a transform at this time.
+    // Due to 2020-07 changes to geometry2, tf buffer will issue warnings for
+    // repeated transforms with the same timestamp
```
MichaelGrupp, line 267: We use full sentences in comments, so please end it with a point character as mentioned by @ojura.

cartographer_ros/cartographer_ros/node.h
```diff
@@ -214,6 +214,7 @@ class Node {
 
   // These are keyed with 'trajectory_id'.
   std::map<int, ::cartographer::mapping::PoseExtrapolator> extrapolators_;
+  std::map<int, ::ros::Time> last_published_tf_stamp_;
```
MichaelGrupp, line 217: Please use plural for consistency with the other maps here. Rename to `last_published_tf_stamps_`.

MichaelGrupp (approved):
Please rebase or click the "Update branch" button here, thanks.

stribor14:
> Please rebase or click the "Update branch" button here, thanks.

Done!

MichaelGrupp (approved):

